PR #1, opened 2016-10-07
Initial checkin.

kmaglione:


browser-polyfill.js
```diff
@@ -0,0 +1,340 @@
...
+     *        {@see wrapAsyncFunction}
+     *
+     * @returns {Proxy<object>}
+     */
+    const wrapObject = (target, wrappers = {}, metadata = {}) => {
+      let cache = Object.create(null);
+
+      let handlers = {
+        has(target, prop) {
+          return prop in target || prop in cache;
```
aswan: how does prop get into the cache if it was never in target?
Ah, perhaps if a script writes a property into chrome or chrome.foo.  Such properties get mirrored to browser or browser.foo?  that seems odd...

browser-polyfill.js
```diff
@@ -0,0 +1,340 @@
...
+              // Promise wrapper for it.
+              let wrapper = wrapAsyncFunction(prop, metadata[prop]);
+              value = wrapMethod(target, target[prop], wrapper);
+            } else {
+              // This is a method that we don't know or care about. Return the
+              // original method, bound to the underlying object.
+              value = value.bind(target);
+            }
+          } else if (typeof value == "object" &&
+                     (hasOwnProperty(wrappers, prop) ||
```
aswan: isn't it possible that wrappers or metadata is undefined here and this throws?

browser-polyfill.js
kmaglione: `foo` winds up in the cache if a script assigns to `browser.foo`. If a script assigns to `chrome.foo`, that winds up visible as `browser.foo`, too, though.

Gruntfile.js
```diff
@@ -0,0 +1,87 @@
...
+          compilation_level: "WHITESPACE_ONLY",
+          language_in: "ECMASCRIPT6_STRICT",
+          language_out: "ECMASCRIPT6",
+          output_wrapper: `${LICENSE}\n%output%`,
+        },
+      },
+    },
+
+    // This currently does not support ES6 classes.
+    uglify: {
```
kmaglione: Not at the moment, since it doesn't currently support ES6. I'm probably going to add a Babel transpilation step and re-enable it, though.

browser-polyfill.js
kmaglione: No, because the arguments have default values.

browser-polyfill.js
```diff
@@ -0,0 +1,340 @@
...
+              value = wrapMethod(target, target[prop], wrappers[prop]);
+            } else if (hasOwnProperty(metadata, prop)) {
+              // This is an async method that we have metadata for. Create a
+              // Promise wrapper for it.
+              let wrapper = wrapAsyncFunction(prop, metadata[prop]);
+              value = wrapMethod(target, target[prop], wrapper);
+            } else {
+              // This is a method that we don't know or care about. Return the
+              // original method, bound to the underlying object.
+              value = value.bind(target);
```
Rob--W: I think that it is better to not blindly forward methods or properties that you do not know. Otherwise it is not possible to support new promisified methods without breaking backwards-compatibility.

browser-polyfill.js
```diff
@@ -0,0 +1,340 @@
...
+            cache[prop] = value;
+          } else {
+            target[prop] = value;
+          }
+          return true;
+        },
+
+        defineProperty(target, prop, desc) {
+          Object.defineProperty(cache, prop, desc);
+          return true;
```
Rob--W: Use `return Reflect.defineProperty(cache, prop, desc);`

In case of non-configurable properties, `Object.defineProperty` throws in strict mode, while you should be returning `false` instead. `Reflect.defineProperty` handles this case just fine.

browser-polyfill.js
```diff
@@ -0,0 +1,340 @@
...
+        },
+
+        deleteProperty(target, prop) {
+          delete cache[prop];
+          return true;
```
Rob--W: Similarly, `return Reflect.deleteProperty(cache, prop);`

browser-polyfill.js
```diff
@@ -0,0 +1,340 @@
...
+     *
+     * @returns {object}
+     */
+    const wrapEvent = wrapperMap => ({
+      addListener(target, listener, ...args) {
+        target.addListener(wrapperMap.get(listener), ...args);
+      },
+
+      hasListener(target, listener) {
+        target.hasListener(wrapperMap.get(listener));
```
Rob--W: You forgot a `return`.

browser-polyfill.js
```diff
@@ -0,0 +1,340 @@
...
+       *        Details about the sender of the message.
+       * @param {function(*)} sendResponse
+       *        A callback which, when called with an arbitrary argument, sends
+       *        that value as a response.
+       * @returns {boolean}
+       *        True if the wrapped listener returned a Promise, which will later
+       *        yield a response. False otherwise.
+       */
+      return function onMessage(message, sender, sendResponse) {
+        let result = listener(message, sender);
```
Rob--W: What if `listener` throws?

rpl:
@kmaglione I took an initial look at the polyfill sources, and, besides the missing return in the `hasListener` wrapper, which Rob as already pointed out, it already looks great!

Follows a small number of nits (and/or other non-mandatory proposed changes):

- exact versions in the npm deps (nit/non-mandatory proposed change): in other of the nodejs projects related to the WebExtensions we are using exact versions in the npm dev dependencies and using greenkeeper to be notified when a new version has been released (and tests runs on travis with the updated dependency before being actually updated in the package.json from the master branch) 

- create a .gitignore file (nit): the repo is currently missing a .gitignore file, it would be nice to create it and add at least "node_modules/" (and "dist/" if the generated files are not going to be committed in the repo)

- docs on how api-metadata.json has been generated (nit): it would be nice to describe somewhere (e.g. the README.md, or maybe a CONTRIBUTING.md?) how the api-metadata.json has been generated (even better if we can include the actual script)

I'm a bit in doubt (but I need to dig a bit more to say if it is actually an issue and in which scenarios) for the exceptions raised in some of the wrappers, in particular I'm a bit concerned that if the polyfill has slightly different behavior from the behavior of our "native" browser API object, the addon code running with the polyfill can potentially follow a different path from when the same code running on Firefox without the polyfill.

rpl:
During this initial digging, I put together a proposal for an initial version of a test suite to unit test the polyfill (in nodejs, using `jsdom`, `mocha`, `chai`, `sinonjs`, included with code coverage reports, using `nyc` and `istanbul-lib-instrument`):

- experimental `proposal/unit-testing` branch compared to the content of this PR: https://github.com/kmaglione/webextension-polyfill/compare/master...rpl:proposal/unit-testing 

- tests running on travis (they currently fail because of the missing return in the hasListener wrapper): https://travis-ci.org/rpl/webextension-polyfill/builds/166448491#L828

I've not yet covered all the polyfill code, but it is missing only a small number of additional test case to be better covered.

browser-polyfill.js
```diff
@@ -0,0 +1,340 @@
...
+            });
+
+            return value;
+          }
+
+          cache[prop] = value;
+          return value;
+        },
+
+        set(target, prop, value, receiver) {
```
rpl: `receiver` is currently unused, how about omitting it from the listed parameters? (or adding an inline comment for it to explain that it is ignored on purpose because we don't need to cover that particular scenario).

browser-polyfill.js
```diff
@@ -0,0 +1,340 @@
...
+     *        An object tree containing metadata used to automatically generate
+     *        Promise-based wrapper functions for asynchronous. Any function in
+     *        the `target` object tree which has a corresponding metadata object
+     *        in the same location in the `metadata` tree is replaced with an
+     *        automatically-generated wrapper function, as described in
+     *        {@see wrapAsyncFunction}
+     *
+     * @returns {Proxy<object>}
+     */
+    const wrapObject = (target, wrappers = {}, metadata = {}) => {
```
rpl: nit

There are a lot of things called `wrapper` and `wrappers` here, how about renaming this `wrappers` parameter into something like "specialAPIWrappers" and the `wrapper` defined in the body into something like `asyncFunctionWrapper`?
(it would make it faster to visually recognize where a wrapper represents a special API or an async function)

browser-polyfill.js
```diff
@@ -0,0 +1,340 @@
...
+            return value;
+          }
+
+          cache[prop] = value;
+          return value;
+        },
+
+        set(target, prop, value, receiver) {
+          if (prop in cache) {
+            cache[prop] = value;
```
rpl: nit

This line is probably not executed currently, if I'm not wrong, it is supposed to be used if there is a setter in the "special API wrappers".

If would be nice to add an inline comment on it to explain when it is supposed to be used (in other word, when the value is going to be forwarded to a cached wrapped setter instead of fowarding it to the related original target setter).

browser-polyfill.js
kmaglione: Then it's dealt with the same way it would normally be dealt with if the listener throws. This version doesn't deal with propagating errors.

browser-polyfill.js
kmaglione: That's not something I'm concerned about at this point.

browser-polyfill.js
```diff
@@ -0,0 +1,340 @@
...
+     */
+    const wrapAsyncFunction = (name, metadata) => {
+      return function asyncFunctionWrapper(target, ...args) {
+        if (args.length < metadata.minArgs) {
+          throw new Error(`Expected at least ${metadata.minArgs} arguments for ${name}(), got ${args.length}`);
+        }
+
+        if (args.length > metadata.maxArgs) {
+          throw new Error(`Expected at most ${metadata.maxArgs} arguments for ${name}(), got ${args.length}`);
+        }
```
kmaglione: In the interest of sensible error messages. We change the number of arguments passed to the underlying function, and the error messages don't make sense when they're complaining about the type of an argument that the user didn't explicitly pass.

browser-polyfill.js
kmaglione: Yes, I'm going to wind up renaming them eventually.

browser-polyfill.js
kmaglione: These are the standard set of arguments passed to proxy hooks. Even if they're not used, they're passed.

browser-polyfill.js
kmaglione: It's executed in a number of circumstances. The semantics aren't very well-defined at the moment, but it only really matters when there are forwarding property descriptors or expandos added by users of the library.

PR #2, opened 2016-10-10
test: introduced a test suite for unit testing.

rpl:
This PR introduces a test suite for unit testing the polyfill inside a fake window context created using jsdom.

The test suite is based on mocha, chai and sinon.

A small hack is used to generate and collect code coverage data:
- the polyfill is instrumented manually (using `istanbul-lib-instrument`)
- the code coverage data is collected and copied from the fake jsdom windw to the node global after every test, where it is retrieved by nyc to generate the code coverage reports


Rob--W:
Random idea: How about running all WebExtensions tests with the `browser` polyfill instead of the native `browser`? That will instantly give much more confidence in the interoperability of the two implementations with little efforts.

kmaglione:
That might be worth doing in the long run, but for now, we're nowhere near being able to run mozilla-central tests against code in Github repos, and we'd need to make a lot of changes to the test harness to make sure the polyfill is injected in all of the contexts that need it.

Either way, we need a separate set of basic unit tests for the polyfill itself.

test/.eslintrc
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": "../.eslintrc",
```
kmaglione: Not necessary.

test/setup.js
```diff
@@ -0,0 +1,75 @@
+"use strict";
+
+const {createInstrumenter} = require("istanbul-lib-instrument");
+const fs = require("fs");
+const {jsdom, createVirtualConsole} = require("jsdom");
```
kmaglione: Please sort imports.

test/setup.js
```diff
@@ -0,0 +1,75 @@
...
+
+function setupTestDOMWindow(chromeObject) {
+  return new Promise((resolve, reject) => {
+    let window;
+
+    // If a jsdom window has already been created, reuse it so that
+    // we can retrieve the final code coverage data, which has been
+    // collected in the jsdom window (where the instrumented browser-polyfill
+    // is running).
+    if (global.window) {
```
kmaglione: If we're going to use a shared global window, can we just create it at the top-level?

test/setup.js
```diff
@@ -0,0 +1,75 @@
...
+      window = global.window;
+      window.browser = undefined;
```
kmaglione: `delete window.browser` would be better.

test/setup.js
```diff
@@ -0,0 +1,75 @@
...
+      const scriptContent = fs.readFileSync(BROWSER_POLYFILL_PATH, "utf-8");
+      scriptEl.textContent = inst.instrumentSync(scriptContent, BROWSER_POLYFILL_PATH);
+    } else {
+      scriptEl.src = BROWSER_POLYFILL_PATH;
+    }
+
+    // Prepare to listen for script loading errors (which results in a rejection),
+    // and to detect when the browser-polyfill has been executed (which resolves
+    // to the jsdom window where the loading has been completed).
+    window.__browserPolyfillLoaded__ = {resolve, reject};
```
kmaglione: Property names bracketed with double underscores are reserved.

Why is this property necessary?

test/setup.js
```diff
@@ -0,0 +1,75 @@
...
+    window.addEventListener("error", (evt) => reject(evt));
+    const loadedScriptEl = window.document.createElement("script");
+    loadedScriptEl.textContent = `
+      window.removeEventListener("error", window.__browserPolyfillLoaded__.reject);
+      window.__browserPolyfillLoaded__.resolve(window);
```
kmaglione: If we're trying to track when the polyfill's script element has finished loading, we should just add "load" and "error" listeners to that element.

test/test-async-functions.js
```diff
@@ -0,0 +1,31 @@
...
+const sinon = require("sinon");
+
+const {setupTestDOMWindow} = require("./setup");
+
+describe("browser-polyfill", () => {
+  describe("wrapped async functions", () => {
+    it("returns a promise which resolves to the callback parameters", () => {
+      const fakeChrome = {
+        alarms: {clear: sinon.stub()},
+        runtime: {lastError: undefined},
```
kmaglione: `null`

test/test-browser-global.js
```diff
@@ -0,0 +1,14 @@
+"use strict";
+
+const {assert} = require("chai");
+
+const {setupTestDOMWindow} = require("./setup");
+
+describe("browser-polyfill", () => {
+  it("automatically wrapps chrome into a browser object", () => {
```
kmaglione: *wraps

Also, probably something like "wraps the global `chrome` namespace with a global `browser` namespace"

test/test-proxied-properties.js
```diff
@@ -0,0 +1,48 @@
...
+        const setResult2 = window.browser.nowrapns.nowrapkey = "new-value";
+
+        assert.equal(setResult, "new-value",
+                     "Got the expected result from setting a wrapped property name");
+        assert.equal(setResult2, "new-value",
+                     "Got the expected result from setting a wrapped property name");
+      });
+    });
+
+    it("delete proxy getter/setter that are not wrapped", () => {
```
kmaglione: *deletes

test/test-runtime-onMessage.js
```diff
@@ -0,0 +1,108 @@
+"use strict";
+
+const {assert} = require("chai");
+const sinon = require("sinon");
+
+const {setupTestDOMWindow} = require("./setup");
+
+describe("browser-polyfill", () => {
+  describe("wrapped runtime.onMessage listener", () => {
+    it("keep track of the listeners added", () => {
```
kmaglione: *keeps

test/test-async-functions.js
```diff
@@ -0,0 +1,31 @@
...
+          .onFirstCall().callsArgWith(1, "res1")
+          .onSecondCall().callsArgWith(1, "res1", "res2", "res3");
+
+        return Promise.all([
+          window.browser.alarms.clear("test1"),
+          window.browser.alarms.clear("test2"),
+        ]);
+      }).then(results => {
+        assert.equal(results[0], "res1", "The first call resolved to a single value");
+        assert.deepEqual(results[1], ["res1", "res2", "res3"],
```
kmaglione: The callback for a single API should be called with the same number of arguments each time, so they should either always be converted to an array, or always passed as a single argument.

test/test-async-functions.js
```diff
@@ -0,0 +1,31 @@
...
+        fakeChrome.alarms.clear
+          .onFirstCall().callsArgWith(1, "res1")
+          .onSecondCall().callsArgWith(1, "res1", "res2", "res3");
+
+        return Promise.all([
+          window.browser.alarms.clear("test1"),
+          window.browser.alarms.clear("test2"),
+        ]);
+      }).then(results => {
+        assert.equal(results[0], "res1", "The first call resolved to a single value");
```
kmaglione: Can we destructure the `assert` methods to globals?

test/test-proxied-properties.js
```diff
@@ -0,0 +1,48 @@
...
+        nowrapns: {nowrapkey: "previous-value"},
+      };
+      return setupTestDOMWindow(fakeChrome).then(window => {
+        const setResult = window.browser.runtime.myprop = "new-value";
+        const setResult2 = window.browser.nowrapns.nowrapkey = "new-value";
+
+        assert.equal(setResult, "new-value",
+                     "Got the expected result from setting a wrapped property name");
+        assert.equal(setResult2, "new-value",
+                     "Got the expected result from setting a wrapped property name");
```
kmaglione: We should check that these are also present in the `fakeChrome` object. And set multiple properties, with different values.

Also, we need to test two separate cases: 1) when we set the property before we ever read it, and 2) when we read the property, set it, and then read it again.

test/test-proxied-properties.js
```diff
@@ -0,0 +1,48 @@
...
+        assert.equal(window.browser.newns.newkey2, "new-value",
+                     "The new non-wrapped getter is cached");
+        assert.deepEqual(setRes, {newkey2: "new-value"},
+                     "Got the expected result from setting a new wrapped property name");
+        assert.deepEqual(window.browser.newns, window.chrome.newns,
+                         "chrome.newns and browser.newns are the same");
+
+
+        delete window.browser.newns.newkey2;
+        assert.equal(window.browser.newns.newkey2, undefined,
```
kmaglione: There should also be a `!(... in ...)` check after `delete`.

test/test-runtime-onMessage.js
```diff
@@ -0,0 +1,108 @@
...
+      const messageListener = sinon.spy();
+
+      const fakeChrome = {
+        runtime: {
+          lastError: undefined,
```
kmaglione: `null`

test/test-runtime-onMessage.js
```diff
@@ -0,0 +1,108 @@
...
+            hasListener: sinon.stub(),
+            removeListener: sinon.spy(),
+          },
+        },
+      };
+
+      return setupTestDOMWindow(fakeChrome).then(window => {
+        fakeChrome.runtime.onMessage.hasListener
+          .onFirstCall().returns(false)
+          .onSecondCall().returns(true);
```
kmaglione: Why?

test/test-runtime-onMessage.js
```diff
@@ -0,0 +1,108 @@
...
+        const wrappedListener = fakeChrome.runtime.onMessage.addListener.firstCall.args[0];
+        wrappedListener("msg", {name: "sender"}, function sendResponse() {});
+        assert.ok(messageListener.calledOnce, "The listener has been called once");
+
+        window.browser.runtime.onMessage.removeListener(messageListener);
+        assert.ok(fakeChrome.runtime.onMessage.removeListener.calledOnce,
+                  "removeListener has been called once");
+        assert.equal(fakeChrome.runtime.onMessage.addListener.secondCall.args[0],
+                     fakeChrome.runtime.onMessage.removeListener.firstCall.args[0],
+                     "both the addListener and removeListenercalls received the same wrapped listener");
```
kmaglione: Check `hasListener` again.

test/test-runtime-onMessage.js
```diff
@@ -0,0 +1,108 @@
...
+      });
+    });
+
+    it("sends the returned value as a message response", () => {
+      const fakeChrome = {
+        runtime: {
+          lastError: undefined,
```
kmaglione: `null`

test/setup.js
rpl: I would prefer to be able to use `delete window.browser`, but unfortunately it didn't worked and I used `window.browser = undefined` as a workaround.

I digged about it and it is a known issue: tmpvar/jsdom#1622

I added an inline comment on it, so that we can remember the "why" of this workaround and when the workaround can be removed.

test/setup.js
rpl: +1
I changed it in the updated version of this test helper.

test/test-runtime-onMessage.js
rpl: the hasListener is a sinon stub, and so it doesn't really track the added listeners:
the first call returns false because the listener has not been added yet, the second call returns true because the listener has been added.

I added some inline comment on this stub definition, so that it is more clear the behavior that we want from this stub (and I added a third call which returns false, to cover the third call received after the listener has been removed)

test/setup.js
rpl: do you mean something like:
```
// Create the jsdom window used to run the tests
const testDOMWindow = jsdom("", {virtualConsole}).defaultView;
global.window = testDOMWindow;
...
function setupTestDOMWindow(chromeObject, browserObject = undefined) {
  ...
}
...
```

if the above is what you were suggesting in the above comment, I applied this change in the last version of this pull request.

test/test-async-functions.js
rpl: @kmaglione I didn't applied this change yet, mostly because I want to be sure that I didn't misunderstood it, do you mean something like:

```
// in every test file:
const {equal, deepEqual} = require("chai").assert;
```

or something like:

```
// in the setup.rb
global.assert = require("chai").assert;

// then in every test file
const {equal, deepEqual} = assert;
```

rpl:
@kmaglione most of the suggested changes from the above review comments should now be integrated in the updated version of this PR
(besides the one from https://github.com/mozilla/webextension-polyfill/pull/2#discussion_r82679577, because I wanted to be sure that I didn't misunderstood the suggestion)

I also added some additional test case, so that the test suite now covers 100% on both lines and branches (obviously, I'm not excluding that there are still test cases that have to be added)

Additional note: Travis is not yet enabled on this repo, it would be nice to enable this repo on Travis asap, so that we can see the travis results on this PR, even before we land it on master.

test/test-async-functions.js
kmaglione: Either is fine with me, but the first is probably better.

test/setup.js
kmaglione: Yes, except that I don't think it needs to be exported to `global.window`.

rpl: ah, yeah! that's true, I removed any `global` usage in the last updates pushed in this pull request

test/test-async-functions.js
rpl: 👍 I prefer the first one too, change applied in the last pushed updates.

(the only reason for the second proposal was that I initially misunderstood what you meant with `global` :-), now I see that clearly it meant to the "module global" and not the "node global global" :-))

test/test-runtime-onMessage.js
rpl: hehehe, sorry :-)

I agree, in this case implementing the "has/add/remove" logic is pretty simple and it makes the test much more readable.

I've applied this change in last pushed updates (in bed17c0), which now wraps with sinon spies a simple "has/add/remove" implementation.

rpl:
@kmaglione thanks for this new round of review comments, I've applied the suggested changes and pushed them in this pull request.

.travis.yml
```diff
@@ -0,0 +1,18 @@
...
+script:
+- npm run build
+- COVERAGE=y npm run test-coverage
+
+after_script: npm run publish-coverage
+
+notifications:
+  irc:
+    channels:
+    - irc.mozilla.org#amo-bots
```
kmaglione: Would probably be better in #webextensions

test/test-async-functions.js
```diff
@@ -0,0 +1,90 @@
...
+
+      return setupTestDOMWindow(fakeChrome).then(window => {
+        // Test for single array callback argument.
+        fakeChrome.tabs.query
+          .onFirstCall().callsArgWith(1, ["res1", "res2"]);
+
+        return window.browser.tabs.query({active: true}).then(
+          () => fail("Expected a rejected promise"),
+          (err) => equal(err, fakeChrome.runtime.lastError,
+                                "Got the expected error in the rejected promise")
```
kmaglione: Please align with opening paren.

test/test-async-functions.js
```diff
@@ -0,0 +1,90 @@
...
+        runtime: {
+          lastError: null,
+          sendMessage: sinon.spy(),
+        },
+      };
+
+      return setupTestDOMWindow(fakeChrome).then(window => {
+        throws(() => {
+          window.browser.runtime.sendMessage();
+        }, "Expected at least 1 arguments for sendMessage(), got 0");
```
kmaglione: Oh. I guess I should add proper pluralization to those messages...

test/test-browser-global.js
```diff
@@ -0,0 +1,65 @@
...
+
+describe("browser-polyfill", () => {
+  it("wraps the global chrome namespace with a global browser namespace", () => {
+    const fakeChrome = {};
+    return setupTestDOMWindow(fakeChrome).then(window => {
+      equal(typeof window.browser, "object", "Got the window.browser object");
+    });
+  });
+
+  it("do not override the global browser namespace if it already exists", () => {
```
kmaglione: s/do/does/

test/test-browser-global.js
```diff
@@ -0,0 +1,65 @@
...
+    const fakeChrome = {
+      runtime: {lastError: null},
+    };
+    const fakeBrowser = {
+      mycustomns: {mykey: true},
+    };
+
+    return setupTestDOMWindow(fakeChrome, fakeBrowser).then(window => {
+      deepEqual(window.browser, fakeBrowser,
+                       "The existent browser has not been wrapped");
```
kmaglione: Please align with opening paren. Also, s/existent/existing/

test/test-browser-global.js
```diff
@@ -0,0 +1,65 @@
...
+      return setupTestDOMWindow(fakeChrome).then(window => {
+        Object.defineProperty(window.browser, "myns", {
+          enumerable: true,
+          configurable: true,
+          value: {mykey: true},
+        });
+
+        ok("myns" in window.browser, "The custom property exists");
+        ok("mykey" in window.browser.myns,
+                  "The content of the custom property exists");
```
kmaglione: Align with paren.

test/test-browser-global.js
```diff
@@ -0,0 +1,65 @@
...
+
+        deepEqual(window.browser.myns, {mykey: true},
+                        "The custom property has the expected content");
```
kmaglione: Align.

test/test-browser-global.js
```diff
@@ -0,0 +1,65 @@
...
+
+        delete window.browser.myns;
+
+        ok(!("myns" in window.browser),
+                  "The deleted custom defined property has been removed");
```
kmaglione: Align.

test/test-browser-global.js
```diff
@@ -0,0 +1,65 @@
...
+      });
+    });
+
+    it("returns undefined for property undefined in the target", () => {
+      const fakeChrome = {myns: {mykey: true}};
+      return setupTestDOMWindow(fakeChrome).then(window => {
+        equal(window.browser.myns.mykey, true,
+                     "Got the expected result from a wrapped property");
```
kmaglione: Align.

test/test-browser-global.js
```diff
@@ -0,0 +1,65 @@
...
+        equal(window.browser.myns.nonexistent, undefined,
+                     "Got undefined for non existent property");
```
kmaglione: Align.

test/test-browser-global.js
```diff
@@ -0,0 +1,65 @@
...
+        equal(window.browser.nonexistent, undefined,
+                     "Got undefined for non existent namespaces");
```
kmaglione: Align.

test/test-proxied-properties.js
```diff
@@ -0,0 +1,104 @@
...
+      return setupTestDOMWindow(fakeChrome).then(window => {
+        ok(window.browser.runtime.nonwrappedmethod);
+
+        const fakeCallback = () => {};
+        window.browser.runtime.nonwrappedmethod(fakeCallback);
+
+        const receivedCallback = fakeChrome.runtime.nonwrappedmethod.firstCall.args[0];
+
+        equal(fakeCallback, receivedCallback,
+                     "The callback has not been wrapped for the nonwrappedmethod");
```
kmaglione: Align.

test/test-proxied-properties.js
```diff
@@ -0,0 +1,104 @@
...
+        runtime: {myprop: "previous-value"},
+        nowrapns: {
+          nowrapkey: "previous-value",
+          nowrapkey2: "previous-value",
+        },
+      };
+      return setupTestDOMWindow(fakeChrome).then(window => {
+        // Check that the property values on the generated wrapper.
+        equal(window.browser.runtime.myprop, "previous-value",
+                     "Got the expected result from setting a wrapped property name");
```
kmaglione: Align

test/test-proxied-properties.js
```diff
@@ -0,0 +1,104 @@
...
+        equal(window.browser.nowrapns.nowrapkey, "previous-value",
+                     "Got the expected result from setting a wrapped property name");
```
kmaglione: Align

test/test-proxied-properties.js
```diff
@@ -0,0 +1,104 @@
...
+
+        // Update the properties on the generated wrapper.
+        const setResult = window.browser.runtime.myprop = "new-value";
+        const setResult2 = window.browser.nowrapns.nowrapkey = "new-value";
+
+        // Check the results of setting the new value of the wrapped properties.
+        equal(setResult, "new-value",
+                     "Got the expected result from setting a wrapped property name");
```
kmaglione: Align.

test/test-proxied-properties.js
```diff
@@ -0,0 +1,104 @@
...
+                     "Got the expected result from setting a new wrapped property name");
+        deepEqual(window.browser.newns, window.chrome.newns,
+                         "chrome.newns and browser.newns are the same");
+
+        delete window.browser.newns.newkey2;
+        equal(window.browser.newns.newkey2, undefined,
+                     "Got the expected result from setting a wrapped property name");
+        ok(!("newkey2" in window.browser.newns),
+                  "The deleted property is not listed anymore");
+      });
```
kmaglione: Eh, you get the picture. Align all of the ones between the last comment and this one. And in the other files.

test/test-runtime-onMessage.js
```diff
@@ -0,0 +1,165 @@
+"use strict";
+
+const {deepEqual, equal, ok} = require("chai").assert;
+const sinon = require("sinon");
+
+const {setupTestDOMWindow} = require("./setup");
+
+describe("browser-polyfill", () => {
+  describe("wrapped runtime.onMessage listener", () => {
+    it("do not wrap the listener if it is not a function", () => {
```
kmaglione: s/do/does/

test/test-runtime-onMessage.js
```diff
@@ -0,0 +1,165 @@
...
+
+        // Add the same listener again to test that it will be called with the
+        // same wrapped listener.
+        window.browser.runtime.onMessage.addListener(messageListener);
+
+        ok(fakeChrome.runtime.onMessage.addListener.calledTwice,
+                  "addListener has been called twice");
+        equal(fakeChrome.runtime.onMessage.addListener.secondCall.args[0],
+                     fakeChrome.runtime.onMessage.addListener.firstCall.args[0],
+                     "both the addListener calls received the same wrapped listener");
```
kmaglione: We should also check that calling with a different function passes through a different wrapper, and so forth.

rpl:
@kmaglione I applied the changes pointed out in the last round of review comments, and I also added the eslint rule that enforces the spacing in the call expressions (and applied the indentation fix using eslint --fix)

I also added a small tweak to pluralize "argument" in the logged messages (and fixed the test accordingly), let me know if it is ok for you or if there is any further change needed.

test/test-runtime-onMessage.js
```diff
@@ -123,41 +123,41 @@ describe("browser-polyfill", () => {
...
         wrappedListener("fake message2", {name: "fake sender2"}, sendResponseSpy);
 
         // Wait the second response promise to be resolved.
         return secondResponse;
       }).then(() => {
         ok(messageListener.calledTwice,
-                  "The unwrapped message listener has been called");
+           "The unwrapped message listener has been called");
         deepEqual(messageListener.secondCall.args,
                          ["fake message2", {name: "fake sender2"}],
```
kmaglione: Indentation is still wrong here. Weird that ESLint is OK with that...

test/test-runtime-onMessage.js
```diff
@@ -123,41 +123,41 @@ describe("browser-polyfill", () => {
...
       }).then(() => {
         wrappedListener("fake message3", {name: "fake sender3"}, sendResponseSpy);
 
         // Wait the third response promise to be rejected.
         return thirdResponse.catch(err => {
           equal(messageListener.callCount, 3,
-                    "The unwrapped message listener has been called");
+                "The unwrapped message listener has been called");
           deepEqual(messageListener.thirdCall.args,
                            ["fake message3", {name: "fake sender3"}],
```
kmaglione: And here.

browser-polyfill.js
```diff
@@ -105,13 +105,15 @@ if (typeof browser === "undefined") {
      *       The generated wrapper function.
      */
     const wrapAsyncFunction = (name, metadata) => {
+      const pluralizeArguments = (numArgs) => numArgs == 1 ? "argument" : "arguments";
```
kmaglione: Thanks!

rpl:
@kmaglione after rebasing this branch on the updated master, I had to apply a small change to the build step in d1b0ff9:
After renaming the source file to "browser-polyfill.in.js", gulp-replace was building it as "dist/browser-polyfill.in.js" instead of "dist/browser-polyfill.js", and then the minify and test steps were not able to find it anymore, the smaller possible fix was to rename it to "src/browser-polyfill.js", so that it would become "dist/browser-polyfill.js" after the gulp-replace step.

rpl:
@kmaglione PR merged, thanks for the review.

PR #8, opened 2016-11-20
Change == to ===

examan:


rpl:
@examan it looks like in this particular context `==` should not introduce any unexpected "type casting" (because on the right we have a number and the parameter that is passed as `numArgs` to the `pluralizeArguments` helper should be a number (it is coming from our json metadata: https://github.com/mozilla/webextension-polyfill/blob/master/api-metadata.json).

is there any issue that you are experiencing using the polyfill that is fixed with this change?

examan:
Sorry, it does't cause any problem. You can just ignore it.

PR #9, opened 2016-11-21
pageAction.show has 1 argument

examan:


rpl:
@examan I just looked into the related JSON schema file (http://searchfox.org/mozilla-central/source/browser/components/extensions/schemas/page_action.json#63) and it seems that you are right:
the first parameter of the `pageAction.show` method is the `tabId` and it doesn't seem to be optional.

I also noticed that the same is true for `pageAction.hide`.

@kmaglione can you take a look at this PR? 
In this repo there isn't a script that generates the api-metadata.json file from the WebExtensions API JSON schema yet, and so I'm not sure how/why these two methods seems to contains the wrong values for `minArgs` / `maxArgs`.

api-metadata.json
```diff
@@ -287,8 +287,8 @@
       "maxArgs": 1
     },
     "show": {
-      "minArgs": 0,
-      "maxArgs": 0
+      "minArgs": 1,
+      "maxArgs": 1
```
kmaglione: This also needs to be changed for the `hide` method.

m4tx:
Actually, shouldn't this be removed completely from the metadata JSON? This PR only fixes the problem with "Expected at most 0 arguments", but the method still doesn't work. The problem is that since the function is inside said JSON, it is wrapped as an asynchronous function - it means that it receives not only the parameters it was called with, but also Promise's resolve function. Of course, it generates an exception in Chrome:
```
Error: Invocation of form pageAction.show(integer, function) doesn't match definition pageAction.show(integer tabId)
```

rpl:
Hi @examan, thanks a lot for looking into this, I've included the changes applied by this PR on the api-metadata.json file as part of #59 (and I took care to preserve you as the author of the commit on that patch).

I'm closing this PR in favor of #59, where we are also working on the additional changes needed to the polyfill wrappers to make it able to return a promise for this APi methods which do not accept a callback on Chrome but returns a promise on Firefox.

Issue #10, opened 2016-11-22
Broken?


Happy-Ferret:
The current master doesn't seem to work.
At least not for me.

When I add the unminimized distribution to ``manifest.json`` (see below), the output is as follows:

``Error: Error: Invocation of form runtime.sendNativeMessage(string, string, function) doesn't match definition runtime.sendNativeMessage(string application, object message, optional function responseCallback)``

utilizing the minimized version, I receive the following error output

``browser-polyfill.min.js:18 Uncaught SyntaxError: Unexpected token {``

An excerpt from my ``manifest.json``

```
"background": {
    "scripts": ["webextension-polyfill/dist/browser-polyfill.js",
                "background.js"]
  },
```

**Browser: Chrome 54.0.2840.99**

rpl:
@Happy-Ferret I took a look into this and it looks like there are two separate issues:

The first one (the "Invocation Error") seems to be related to a small difference in the definitions of this API method in the Firefox and Chrome JSON schema files:

- on Firefox this method definition contains 3 parameters: a native application name with a string type, a message with type any and the optional callback (http://searchfox.org/mozilla-central/source/toolkit/components/extensions/schemas/runtime.json#374-378)
- on Chrome this method definition contains the same 3 parameters, but the message is "type object" (https://cs.chromium.org/chromium/src/extensions/common/api/runtime.json?dr=C&q=sendNativeMessage+file:json$&sq=package:chromium&l=366-373)

Given the above schema API method definitions, an extension that wants to be compatible with both Chrome and Firefox have to use a message of type object (which is supported by both the definitions), e.g.:

```js
browser.runtime.sendNativeMessage("ping_pong", {data: "a string message in an object"})
  .then(reply => console.log("Native message reply", reply))
  .catch(err => console.error("Native message error", err));
```

Let me know it the above workaround fixes the "Invocation Error" for you too.

I'm still looking into the minified file issue (the "SyntaxError"), it looks like like something that can be related to some ES6 syntax not correctly transformed by the additional "minification" step.

rpl:
@Happy-Ferret I looked deeper into the "minified file's `SyntaxError`" part of this issue and it looks like that my initial guess was right: google-closure-compiler seems to be the one that introduces the issue when the sources are minified.

I did an brief experiment by using [babili](https://github.com/babel/babili) as an alternative minifier that is officially "ES6-aware" and I created a pull request for it in #11.

[...]

PR #11, opened 2016-11-22
fix: use babili to minify ES6 sources.

rpl:
This PR introduces the changes needed to use [babili](https://github.com/babel/babili) instead of google-closure-compiler to generate the minified version of the original ES6 sources.

The goal of this PR is producing a minified version of the `webextension-polyfill` sources that can be loaded successfully on Chrome (instead of raising a "SyntaxError" exception as described in #10)

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -8,6 +8,8 @@ const LICENSE = `/* This Source Code Form is subject to the terms of the Mozilla
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */`;
 
+const MINIFIED_FILE_FOOTER = `\n\n// <%= pkg.name %> v.<%= pkg.version %> (<%= pkg.homepage %>)\n\n${LICENSE}`;
+
 module.exports = function(grunt) {
   grunt.initConfig({
     pkg: grunt.file.readJSON("package.json"),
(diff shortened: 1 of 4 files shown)


rpl:
@kmaglione can you take a look at this change?

During #10 debugging, I noticed that the "SyntaxError" raised by the minified file seems to be related to how google-closure-compiler minifies the ES6 sources, this PR uses babili (which is based on babel) and it seems to be able to generate a working minified version of the polyfill.

Gruntfile.js
```diff
@@ -55,41 +57,26 @@ module.exports = function(grunt) {
...
-          // Closure currently supports only whitespace and comment stripping
-          // when both the input and output languages are ES6.
-          compilation_level: "WHITESPACE_ONLY",
-          language_in: "ECMASCRIPT6_STRICT",
-          language_out: "ECMASCRIPT6",
-          output_wrapper: `${LICENSE}\n%output%`,
-        },
+    shell: {
+      minify: {
+        command: "babili dist/browser-polyfill.js -o dist/browser-polyfill.min.js --minified --no-comments --source-maps true",
```
kmaglione: Please use grunt-babel for this instead, and ideally set up a minify env in `package.json`

Gruntfile.js
```diff
@@ -55,41 +57,26 @@ module.exports = function(grunt) {
...
-
-      dist: {
-        files: {
-          "dist/browser-polyfill.min.js": ["dist/browser-polyfill.js"],
-        },
+    concat: {
+      license: {
+        src: "dist/browser-polyfill.min.js",
+        dest: "dist/browser-polyfill.min.js",
+        options: {footer: MINIFIED_FILE_FOOTER},
```
kmaglione: Please do this via babel using `auxiliaryCommentBefore`.

Gruntfile.js
rpl: :+1: 

patch updated to use grunt-babel and the babeli preset.

Gruntfile.js
rpl: Unfortunately there are two issues that I noticed by trying to use `auxiliaryCommentBefore` as suggested above:

- it didn't work if the babel is configured to do not output any comment (which is needed to exclude all the inline comments from the minified version)
 
- it didn't work as I expected initially (it doesn't add a single comment, but multiple comments, basically its "most common use case" is: including the "istanbul ignore next" on the code injected by babel that is not coming from the original transpiled code)

rpl:
@kmaglione I also added a proposed additional step in the travis.yml which is going to re-run all the unit tests on the minified file (mostly because it already happened that the minified version was broken, which is the reason of this proposed change, and babili is a pretty young project and we cannot completely exclude that it could happen again)

rpl:
Merged as 9a38d923 and 96cfb9c5

PR #13, opened 2016-12-27
Remove pageAction.show and .hide from API metadata

m4tx:
Their presence there caused them to be wrapped as asynchronous functions. It caused Promise's resolve function to be passed to them, which generated error (`Invocation of form pageAction.show(function) doesn't match definition pageAction.show(integer tabId)`).

This is a response to PR #9.

### DIFF
--- a/api-metadata.json
+++ b/api-metadata.json
@@ -278,17 +278,9 @@
       "minArgs": 1,
       "maxArgs": 1
     },
-    "hide": {
-      "minArgs": 0,
-      "maxArgs": 0
-    },
     "setIcon": {
       "minArgs": 1,
       "maxArgs": 1
-    },
-    "show": {
-      "minArgs": 0,
-      "maxArgs": 0
     }
   },
   "runtime": {


Rob--W:
FYI: The fact that some pageAction/browserAction methods in Chrome do not take callbacks is a bug in Chrome. I expect Chrome to eventually support callbacks for these APIs:

- https://bugs.chromium.org/p/chromium/issues/detail?id=366282#c8
- https://bugs.chromium.org/p/chromium/issues/detail?id=451320#c2

kmaglione:
Sorry for the delay.

These methods need to stay in the polyfill, but in Chrome they should just return an immediately resolved Promise rather than passing a callback.

Rob--W:
> These methods need to stay in the polyfill, but in Chrome they should just return an immediately resolved Promise rather than passing a callback.

As said before, I expect the callback to be implemented eventually, so doing something like `if (browser is Chrome) ...` wouldn't fly well. The logic should be something like this:

```javascript
new Promise((resolve, reject) => {
    try {
        chrome.pageAction.show(tabId, function() {
            if (chrome.runtime.lastError) reject(chrome.runtime.lastError);
            else resolve();
        });
    } catch (e) { // Error. Callback argument is probably not supported.
        chrome.pageAction.show(tabId);
        resolve();
    }
});
```

rpl:
Closing as *state: invalid* based on comment https://github.com/mozilla/webextension-polyfill/pull/13#issuecomment-277090011 (we are going to fix this issue by returning a resolved promise based on an updated `"api-metadata.json"` file)

Issue #14, opened 2017-01-12
ESLint causes build on windows to fail with default git settings


rmtmckenzie:
Because ESLINT is set to need linefeeds set to LF, by default on windows building this will fail.

There's a few possible workarounds - the easiest being setting global git config core.autocrlf to force linefeeds to LF. However, this can break other programs' builds. A local git config can also be used but that doesn't help if using NPM to pull this repo.

I'm going to do a pull request for adding a .gitattributes file which forces line feeds to LF regardless of the git config for *.js files in the repo.

T-vK:
All you need to do is edit the .eslintrc file. Change `"linebreak-style": [2, "unix"],` to `"linebreak-style": [2, "windows"],` when you are on Windows.

Rob--W:
This was fixed by #103.

PR #15, opened 2017-01-12
Force line endings to LF for js files

rmtmckenzie:
Fixes build on windows by allowing ESLINT to pass when core.autolf is set to anything other than forcing LF (as is the default for windows).

Fixes issue #14


Rob--W:
This already landed in #103.

Issue #7, opened 2016-11-15
Offer an easier way to upgrade the polyfill in a WebExtension


kumar303:
The documentation currently suggests this approach for installing the polyfill in a WebExtension:

* Clone the `webextension-polyfill` repo
* Build the polyfill distribution
* Copy the distribution into your WebExtension source

This works fine the first time but it doesn't offer an easy way to upgrade the polyfill when the upstream source changes.

Can you document (or implement) a better way? There would be a couple ways to approach it. The first thing that comes to mind is this:

* Add a `package.json` to the WebExtension
* Declare `webextension-polyfill` as a dependency
* Configure the WebExtension project with babel to allow imports from `node_modules` and build an es6-friendly distribution for your WebExtension source files
* Every time the dependencies are upgraded, the polyfill will be upgraded

Perhaps there is a simpler way that does not involve babel transpilation. Maybe with [bower](https://bower.io/) ?

ivantsov:
Why don't publish this project as an npm package? It would solve lots of problems.

fiveNinePlusR:
I would like this as well. Is there something technical that would stop this from happening?

josephfrazier:
See https://github.com/mozilla/webextension-polyfill/pull/17 for a patch that makes it possible to `npm install` from github and use with a bundler. If it's merged and subsequently published to npm, it will also be possible to `npm install` the pre-built `dist/` files, so that no build tools or bundlers are needed.

---

If you'd like, you can go ahead and install it from github using:

    npm install --save https://github.com/mozilla/webextension-polyfill#68ee5bfa97fa2b81d9bf2805488872dffb37c15b

Purexo:
Adding build in github release could be nice ;-)

So anyone can dl the build and place it in his project.

Issue #5, opened 2016-11-07
Prevent accidentally using a source version of the polyfill


kumar303:
In https://github.com/mdn/webextensions-examples/pull/131 the developer accidentally included the polyfill source file instead of a built source file. RTFM'ing will prevent this from happening in most cases but could we also prevent it in another way? I think it will cut down on invalid bug reports. Ideas:

* log an error that says "this is the wrong file" (or something) if you run the polyfill from source file.
* ... or completely die with an exception when polyfilling from source
* figure out how to strip these errors out in the dist file

jeffesquivels:
Hi,

Guilty as charged (I'm the developer of  mdn/webextensions-examples#131) :grin: .

My five cents: the main source of confusion was that the source file seemed to work fine initially (i.e. I could use the browser namespace in Chrome), so I guess if there is an easy way to make the source file not work at all, that would be good but I also don't think this issue would happen very often (as most extension developers would probably have more experience than me with the JS ecosytem ;-)), plus the changes introduced in commit c8826128550e8d85658460b22318061814206bf4 should be enough to make it clear, at least for me, that the file in the repo is a source file.

Kind regards,

rpl:
This should now be even more explicit:
 starting from [31d778c2](https://github.com/mozilla/webextension-polyfill/commit/31d778c2cde2b15843fbecc5c18392ddb4c2948e#diff-a8dc0200482fd536ca439ed8bcf763c2R22), the polyfill raises an explicit error (Currently "api-metadata.json has not been included in browser-polyfill") when the source file has been wrongly included instead of the built one.

PR #17, opened 2017-01-26
Add CommonJS export to package.json

josephfrazier:
This addresses [issue 7] by making it possible for users to run:

```sh
npm install https://github.com/mozilla/webextension-polyfill
```

and download a module that they can use with a bundler as follows:

```js
import browser from 'webextension-polyfill';
```

Also, add a [prepublish script] so that users don't need to run `grunt`
manually. In addition, specify [files] in package.json so that this
module can be published to npm without including miscellanea. This can
be verified by running:

```sh
npm pack && tar -tvf webextension-polyfill-0.1.0.tgz
```

The `dist/` files would be generated upon publishing the package,
meaning that users could run `npm install webextension-polyfill`, then
load the following files into their extension, without having to build
them:

```
node_modules/webextension-polyfill/dist/browser-polyfill.js
node_modules/webextension-polyfill/dist/browser-polyfill.min.js
```

[issue 7]: https://github.com/mozilla/webextension-polyfill/issues/7
[files]: https://docs.npmjs.com/files/package.json#files
[prepublish script]: https://docs.npmjs.com/misc/scripts

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,6 @@ simply run:
 
 ```sh
 npm install
-grunt
 ```
 
 This will build both non-minified and minified versions of the final library,
--- a/src/browser-polyfill.js
+++ b/src/browser-polyfill.js
@@ -6,8 +6,6 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 "use strict";
 
-var browserPolyfill; // eslint-disable-line no-unused-vars
-
 if (typeof browser === "undefined") {
   // Wrapping the bulk of this polyfill in a one-time-use function is a minor
   // optimization for Firefox. Since Spidermonkey does not fully parse the
@@ -15,7 +13,7 @@ if (typeof browser === "undefined") {
   // never actually need to be called, this allows the polyfill to be included
   // in Firefox nearly for free.
   const wrapAPIs = () => {
-    const apiMetadata = {/* include("api-metadata.json") */};
+    const apiMetadata = require("../api-metadata.json");
 
     /**
      * A WeakMap subclass which creates and stores a value for any key which does
@@ -338,7 +336,7 @@ if (typeof browser === "undefined") {
     return wrapObject(chrome, staticWrappers, apiMetadata);
   };
 
-  browserPolyfill = wrapAPIs();
+  module.exports = wrapAPIs();
 } else {
-  browserPolyfill = browser;
+  module.exports = browser;
 }
--- a/src/browser-polyfill.js
+++ b/src/browser-polyfill.js
@@ -6,6 +6,8 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 "use strict";
 
+var browserPolyfill; // eslint-disable-line no-unused-vars
+
 if (typeof browser === "undefined") {
   // Wrapping the bulk of this polyfill in a one-time-use function is a minor
   // optimization for Firefox. Since Spidermonkey does not fully parse the
@@ -336,5 +338,7 @@ if (typeof browser === "undefined") {
     return wrapObject(chrome, staticWrappers, apiMetadata);
   };
 
-  this.browser = wrapAPIs();
+  browserPolyfill = wrapAPIs();
+} else {
+  browserPolyfill = browser;
 }
--- a/src/browser-polyfill.js
+++ b/src/browser-polyfill.js
@@ -13,7 +13,9 @@ if (typeof browser === "undefined") {
   // never actually need to be called, this allows the polyfill to be included
   // in Firefox nearly for free.
   const wrapAPIs = () => {
-    const apiMetadata = require("../api-metadata.json");
+    // Note that `require` does NOT work in general. See discussion here:
+    // https://github.com/mozilla/webextension-polyfill/pull/17#discussion_r99170958
+    const apiMetadata = require("../api-metadata.json"); // eslint-disable-line no-undef
 
     /**
      * A WeakMap subclass which creates and stores a value for any key which does
@@ -336,7 +338,9 @@ if (typeof browser === "undefined") {
     return wrapObject(chrome, staticWrappers, apiMetadata);
   };
 
-  module.exports = wrapAPIs();
+  // The build process adds a UMD wrapper around this file, which makes the
+  // `module` variable available.
+  module.exports = wrapAPIs(); // eslint-disable-line no-undef
 } else {
-  module.exports = browser;
+  module.exports = browser; // eslint-disable-line no-undef
 }
--- a/.eslintrc
+++ b/.eslintrc
@@ -7,6 +7,7 @@
     "browser": true,
     "es6": true,
     "webextensions": true,
+    "commonjs": true,
   },
 
   "globals": {
--- a/.eslintrc
+++ b/.eslintrc
@@ -7,7 +7,6 @@
     "browser": true,
     "es6": true,
     "webextensions": true,
-    "commonjs": true,
   },
 
   "globals": {
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -28,7 +28,7 @@ module.exports = function(grunt) {
         options: {
           patterns: [
             {
-              match: /\{\/\* include\("(.*?)"\) \*\/\}/,
+              match: /require\("..\/(.*?)"\)/,
               replacement: (match, filename) => {
                 return grunt.file.read(filename)
                             .replace(/\n$/, "")
@@ -58,7 +58,7 @@ module.exports = function(grunt) {
     umd: {
       all: {
         src: "dist/browser-polyfill.js",
-        objectToExport: "browserPolyfill",
+        template: "unit",
         globalAlias: "browser",
         amdModuleId: "webextension-polyfill",
       },
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -55,6 +55,15 @@ module.exports = function(grunt) {
       },
     },
 
+    umd: {
+      all: {
+        src: "dist/browser-polyfill.js",
+        objectToExport: "browserPolyfill",
+        globalAlias: "browser",
+        amdModuleId: "webextension-polyfill",
+      },
+    },
+
     "closure-compiler": {
       dist: {
         files: {
@@ -88,8 +97,9 @@ module.exports = function(grunt) {
 
   grunt.loadNpmTasks("gruntify-eslint");
   grunt.loadNpmTasks("grunt-replace");
+  grunt.loadNpmTasks("grunt-umd");
   grunt.loadNpmTasks("grunt-coveralls");
   require("google-closure-compiler").grunt(grunt);
 
-  grunt.registerTask("default", ["eslint", "replace", "closure-compiler"]);
+  grunt.registerTask("default", ["eslint", "replace", "umd", "closure-compiler"]);
 };
--- a/package.json
+++ b/package.json
@@ -2,6 +2,10 @@
   "name": "webextension-polyfill",
   "version": "0.1.0",
   "description": "A lightweight polyfill library for Promise-based WebExtension APIs in Chrome.",
+  "main": "dist/browser-polyfill.js",
+  "files": [
+    "dist"
+  ],
   "repository": {
     "type": "git",
     "url": "git+https://github.com/mozilla/webextension-polyfill.git"
@@ -18,6 +22,7 @@
     "grunt": "^1.0.1",
     "grunt-coveralls": "^1.0.1",
     "grunt-replace": "*",
+    "grunt-umd": "^2.4.0",
     "gruntify-eslint": "*",
     "eslint": "3.9.1",
     "istanbul-lib-instrument": "^1.1.3",
@@ -34,6 +39,7 @@
     "build": "grunt",
     "test": "mocha",
     "test-coverage": "COVERAGE=y nyc mocha",
-    "publish-coverage": "grunt coveralls"
+    "publish-coverage": "grunt coveralls",
+    "prepublish": "npm run build"
   }
 }
--- a/package.json
+++ b/package.json
@@ -2,8 +2,10 @@
   "name": "webextension-polyfill",
   "version": "0.1.0",
   "description": "A lightweight polyfill library for Promise-based WebExtension APIs in Chrome.",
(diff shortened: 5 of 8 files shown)


rpl:
Hi @josephfrazier! First of all thanks for looking into this!

Provide the webextension-polyfill as an npm package is definitely something that we should do, what concerns me a bit about how this PR achieves this goal is that the browserify loader is going to be always included in the final built file, even when the addon that includes this library is not using browserify (or any other module loader).

How about trying to wrap it using the UMD format instead?

By wrapping it in an UMD format (Universal Module Format), no browserify loader will be included in final javascript file, nevertheless it would be loadable using a simple tag script, an AMD module loader or a CommonJS module loader (e.g. browserify, webpack etc.)

As an example, it seems that the `grunt-umd` would be able to wrap our library as described above, 
the changes needed to integrate are not not far from the ones that you are already using with browserify in this pull request, and the final javascript will be generated from the pipeline that follows:

- injecting the json file into the initial source file of the polyfill (dist/browser-polyfill.js)
- wrapping the sources into an UMD module wrapper (which overwrites dist/browser-polyfill.js) 
- generate a minified version of the final UMD module (which generates dist/browser-polyfill.min.js)

I've briefly tried `grunt-umd` locally and it seems that the following changes are enough to integrate it:

- load and configure grunt-umd in the Gruntfile

```diff
diff --git a/Gruntfile.js b/Gruntfile.js
index 7f6e503..6d2929b 100644
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -55,6 +55,15 @@ module.exports = function(grunt) {
       },
     },

+    umd: {
+      all: {
+        src: "dist/browser-polyfill.js",
+        objectToExport: "browserPolyfill",
+        globalAlias: "browser",
+        amdModuleId: "webextension-polyfill",
+      },
+    },
+
     "closure-compiler": {
       dist: {
         files: {
@@ -88,8 +97,9 @@ module.exports = function(grunt) {

   grunt.loadNpmTasks("gruntify-eslint");
   grunt.loadNpmTasks("grunt-replace");
+  grunt.loadNpmTasks("grunt-umd");
   grunt.loadNpmTasks("grunt-coveralls");
   require("google-closure-compiler").grunt(grunt);

-  grunt.registerTask("default", ["eslint", "replace", "closure-compiler"]);
+  grunt.registerTask("default", ["eslint", "replace", "umd", "closure-compiler"]);
 }; 
```

- small tweaks on the javascript sources (similar to the one that you are currently applying in this PR)

```diff
diff --git a/src/browser-polyfill.js b/src/browser-polyfill.js
index b7041e7..2cbb02e 100644
--- a/src/browser-polyfill.js
+++ b/src/browser-polyfill.js
@@ -6,6 +6,8 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 "use strict";

+var browserPolyfill; // eslint-disable-line no-unused-vars
+
 if (typeof browser === "undefined") {
   // Wrapping the bulk of this polyfill in a one-time-use function is a minor
   // optimization for Firefox. Since Spidermonkey does not fully parse the
@@ -336,5 +338,7 @@ if (typeof browser === "undefined") {
     return wrapObject(chrome, staticWrappers, apiMetadata);
   };

-  this.browser = wrapAPIs();
+  browserPolyfill = wrapAPIs();
+} else {
+  browserPolyfill = browser;
 }
```

- small changes to the package.json (which are almost the same ones that you are applying in this PR)

Let me know what you think about this alternative strategy.

josephfrazier:
Thanks for the comprehensive feedback, @rpl! I definitely understand the concern about having `browserify` boilerplate in the built file, so I'll push some changes that use `grunt-umd` instead.

src/browser-polyfill.js
```diff
@@ -336,5 +336,7 @@ if (typeof browser === "undefined") {
     return wrapObject(chrome, staticWrappers, apiMetadata);
   };
 
-  this.browser = wrapAPIs();
+  module.exports = wrapAPIs();
```
rpl: I mostly like that by using the `unit` grunt-umd template we didn't need to rename the global alias to be able to export it, but I'm a bit worried that we have changed the eslint config to enable the "commonjs" mode, which can detect as "valid" something that is valid in a regular commonjs module but that we probably want to be mentioned as errors (e.g. usage of the any usage of the `require` or other globals available in a nodejs module).

How about reverting the change to `.eslintrc` and opt to use `// eslint-disable-line no-undef` on these two lines?

something like:
```diff
diff --git a/src/browser-polyfill.js b/src/browser-polyfill.js
index 5d55403..f1ab55c 100644
--- a/src/browser-polyfill.js
+++ b/src/browser-polyfill.js
@@ -336,7 +336,7 @@ if (typeof browser === "undefined") {
     return wrapObject(chrome, staticWrappers, apiMetadata);
   };

-  module.exports = wrapAPIs();
+  module.exports = wrapAPIs(); // eslint-disable-line no-undef
 } else {
-  module.exports = browser;
+  module.exports = browser; // eslint-disable-line no-undef
 }
```

If we opt for this solution (grunt-umd 'unit' template and module.exports) instead of renaming the global as I was suggesting in my previous comment, I would also add an inline comment near to this code, to explain that these `module.exports` are defined in the UMD wrapper.

src/browser-polyfill.js
```diff
@@ -13,7 +13,7 @@ if (typeof browser === "undefined") {
   // never actually need to be called, this allows the polyfill to be included
   // in Firefox nearly for free.
   const wrapAPIs = () => {
-    const apiMetadata = {/* include("api-metadata.json") */};
+    const apiMetadata = require("../api-metadata.json");
```
josephfrazier: > My guess is that you are preferring the require here so that the github repo can be used as an npm dependencies with browserify/webpack without the need for a build step.

That's a side benefit, yes, but my goal this time was actually to make it so that this package can be `import`ed or `require()`ed by other projects without including UMD boilerplate, while build process still produces a UMD module for use in a script tag.

> But I'm a bit concerned that by looking at the sources (without looking at the building process) this require can be misinterpreted

I definitely agree here, and even felt a little dirty writing the `grunt-replace` rule for the `require()` syntax (especially since it has to strip out the leading `../` in the path) 😝 

However, that I think about it, there is a nice incidental effect of having it this way: It makes it much harder for people to accidentally use the source version of the polyfill (which happened in https://github.com/mozilla/webextension-polyfill/issues/5)

Ultimately, though, I'm not really devoted to this syntax, and would happily switch back to `{/* include("...") */}` if that's what the team would prefer.

README.md
```diff
@@ -14,7 +14,6 @@ simply run:
 
 ```sh
 npm install
-grunt
```
josephfrazier: Since `npm run build` was added to package.json as the `prepublish` script, it will run as part of the `npm install` lifecycle. Granted, this behavior is deprecated and will be removed in `npm` v5 (see [here](https://github.com/npm/npm/blob/30d75e738b9cb7a6a3f9b50e971adcbe63458ed3/doc/misc/npm-scripts.md#prepublish-and-prepare) for details), so this part of the documentation will need to be changed anyway.

What if I changed the line to this?

```sh
npm run build # if you're using npm v5+
```

src/browser-polyfill.js
josephfrazier: Ooh, that's a good point about "commonjs" mode causing eslint to be too lenient. I've pushed up https://github.com/mozilla/webextension-polyfill/pull/17/commits/35e8e7a465256945367c16216508593e6024837e with the suggested changes. Of course, if we go back to renaming the global, this will have to be reverted/etc.

josephfrazier:
Cheers for the comments, @rpl. I've responded and I look forward to seeing where the discussion goes.

As for the bundler/AMD testing: I took a very quick shot at testing browserify/webpack in Node (without integrating it into the main test suite). I don't know if this is a good way to go about it, but you can at least have a look at it here: https://github.com/josephfrazier/webextension-polyfill/commit/f35f23a3e969d0d32a794f3126df030ef5a12bc4

README.md
rpl: how about defining both `prepublish` and `prepare`? If I read the deprecation note correctly, it would make the package able to work correctly across the npm versions, am I right?

src/browser-polyfill.js
rpl: But the `require` will not work correctly if the source version is loaded accidentally instead of the build version as a plain tag script added to an extension page, or as an AMD module.

We can prevent the "CommonJS" loader (and bundler) to pick the wrong file by changing the "main" property in the package.json to the dist file.

Let's wait a feedback from @kmaglione on this (but I vote for switching back to the `{/* include("...") */}` ;-)

rpl:
@josephfrazier Thanks for the updates, for the detailed comments and also for starting to look into the test case!

I'm not sure that it is really necessary to test the "commonjs bundled module"-mode with both browserify and webpack (there should be no difference between the two tools from this "package point of view", also I didn't tried yet but we can probably test it even using directly the nodejs loader and pre-populating the node [global](https://nodejs.org/api/globals.html#globals_global) with the expected "chrome" global)

Another additional scenario that we need to add a test for is  the "AMD module"-mode using RequireJS.

I would be also better to integrate the new tests into the test suite.

In the meantime, I'm assigning this review to Kris to get the final feedback on the currently debated points from this change.

README.md
rpl: @josephfrazier we should also take into account that the `devDependencies` are not retrieved when the package is installed as a dependency of another one, and so I'm not entirely sure that defining `prepublish`/`prepare` will be enough to get the the dist files correctly built in this scenario.

josephfrazier: > how about defining both `prepublish` and `prepare`? If I read the deprecation note correctly, it would make the package able to work correctly across the npm versions, am I right?

I just tried defining both `prepublish` and `prepare`, and it causes the build to happen twice when `npm install` is run using npm v4.  So, it would work correctly across npm versions, but just takes a little extra time on npm v4, whereas npm v3 runs the build only once (as will npm v5 once it's released).

We can fix this by using [in-publish](https://www.npmjs.com/package/in-publish) to prevent the `prepublish` script from running upon `npm install` with npm v4. Alternatively, if all the developers who will be publishing this are already using npm v4, we can change `prepublish` to `prepublishOnly`, as detailed in the deprecation note linked in my previous comment.

> we should also take into account that the `devDependencies` are not retrieved when the package is installed as a dependency of another one, and so I'm not entirely sure that defining `prepublish`/`prepare` will be enough to get the the dist files correctly built in this scenario

Note that none of these scripts (`prepublish`, `prepare`, `prepublishOnly`) would be invoked when a user runs `npm install webextension-polyfill` (which is different from running `npm install` inside a clone of this repo). Since running `npm publish` invokes the `prepublish` script, the dist/ files will get built and included in the published package, so a user who runs `npm install webextension-polyfill` will have them without needing to install the `devDependencies`.

As I mentioned when I opened the PR, you can verify this by running the following:

```sh
npm pack && tar -tvf webextension-polyfill-0.1.0.tgz
```

src/browser-polyfill.js
josephfrazier: > But the `require` will not work correctly if the source version is loaded accidentally instead of the build version as a plain tag script added to an extension page, or as an AMD module.

Yes, that's the point :) Instead of silently failing because the following line is valid JS, there would be an `Uncaught ReferenceError`. Well, if both `module` and `require` are defined globally, there could still be a silent failure, but this seems unlikely.

```js
    const apiMetadata = {/* include("api-metadata.json") */};
```

> We can prevent the "CommonJS" loader (and bundler) to pick the wrong file by changing the "main" property in the package.json to the dist file.

Maybe we're misunderstanding each other here, but bundlers like browserify and webpack automatically make sure that `require`/`module`/etc are defined correctly, so they should use `src/browser-polyfill.js`. Of course, if a user wanted to `npm install webextension-polyfill`, then use a script tag to include the polyfill, they could be slightly misled by seeing `"main": "src/browser-polyfill.js"` in package.json, but I think that can be easily handled in documentation.

> Let's wait a feedback from @kmaglione on this (but I vote for switching back to the {/* include("...") */} ;-)

Agreed on waiting for feedback :) Another way to go about this would be to switch back to `{/* include("...") */}`, but then change it to something that throws an error if it's not replaced. That would also make it more obvious to the user if they accidentally use the src file in a script tag.

josephfrazier:
> I'm not sure that it is really necessary to test the "commonjs bundled module"-mode with both browserify and webpack (there should be no difference between the two tools from this "package point of view", also I didn't tried yet but we can probably test it even using directly the nodejs loader and pre-populating the node global with the expected "chrome" global)

Good point, we should assume that bundlers work properly, and instead test that their input is correct (like you're suggesting with nodejs tests). I gave this a shot and pushed up [commits dd0c355 and 99faff9](https://github.com/mozilla/webextension-polyfill/compare/35e8e7a465256945367c16216508593e6024837e...99faff94ffbf03094361875a10e1692acc4b8662) to show you all what I came up with. Let me know if it looks like the right direction.

> Another additional scenario that we need to add a test for is the "AMD module"-mode using RequireJS.

I've thought a little bit about this, but I'm not too familiar with AMD, so I'm not sure if it makes more sense to try to integrate AMD tests with the jsdom tests, or if it would be easier to make more node-style tests like the above, but use the `amdefine` or `requirejs` packages to allow AMD to work in Node. See here for more info: http://requirejs.org/docs/node.html

rpl:
Merged as f9248e62 (with minor tweaks applied by 31d778c2, mostly the same minor changes described in the comments from the last review).

rpl:
@josephfrazier Thank you so much for this, and my sincere apologies for the long waiting time.

During the last week, Kris and I have been finally able to have a brief chat about this PR, and we agreed on the changes to apply on it before landing it in the master branch.

Given how much time you have been waiting, I opted to apply the needed changes on top of your patches as 31d778c2, while I squashed all the patches from this PR in a single commit (f9248e62), with the git metadata preserved (in particular the author, you ;-), and the commit description).

josephfrazier:
@rpl Glad to see this finally land! Since it's still not possible to install directly from github, I took the liberty of publishing the current master branch (96cfb9c5c260053141458d1fa227c78aa29bebf3) to NPM as [webextension-polyfill](https://www.npmjs.com/package/webextension-polyfill). I also added [you](https://www.npmjs.com/~rpl) as an owner of the package, so feel free to take it over from here :)

PR #19, opened 2017-02-02
test: add linting to test scripts

josephfrazier:
Making `grunt eslint` the `pretest` script causes it to run as part of
`npm test`. Likewise, making the `test-coverage` script run `npm test`
(instead of running `mocha` directly) ensures that the Travis builds
will run eslint as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,8 +32,9 @@
   },
   "scripts": {
     "build": "grunt",
+    "pretest": "grunt eslint",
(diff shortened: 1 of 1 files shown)


rpl:
This pull requests has currently conflicts, and we run eslint on travis (and so eslint errors should not going to be undetected), we would like to remove grunt completely at some point but that is not something that we are going to do for now.

PR #22, opened 2017-02-08
Support sendResponse callback in onMessage

### LINKED ISSUE #16: runtime.onMessage is broken
```
Error in event handler for runtime.onMessage: TypeError: sendResponse is not a function
    at runtime.onMessage.addListener (chrome-extension://oommdnlanbpnfobibbnfmnmobfddkofp/background.js:4:2)
    at onMessage (chrome-extension://oommdnlanbpnfobibbnfmnmobfddkofp/browser-polyfill.js:1338:22)
```
https://github.com/mozilla/webextension-polyfill/blob/master/src/browser-polyfill.js#L314-L328
I'm not sure this is due to API changes or something. The wrapper of listener looks completely nonsense to me. Shouldn't it just invoke the listener directly?
```
      return function onMessage(message, sender, sendResponse) {
        return listener(message, sender, sendResponse);
      };
```

Rob--W:
Now onMessage behaves as documented in MDN, and like Firefox.

The only thing I was not completely sure about is what to do when `return true` is used. It can be resolved in two ways:

1. Calling `sendResponse` with `true`.
2. Returning `true` in the real `onMessage` handler and allowing `sendResponse` to be called asynchronously.

I follow Firefox's implementation and do the latter - http://searchfox.org/mozilla-central/rev/848c29538ab007fb95dc6cff194f0e3d3809613d/toolkit/components/extensions/ExtensionChild.jsm#363

This fixes #16.

### DIFF
--- a/src/browser-polyfill.js
+++ b/src/browser-polyfill.js
@@ -312,7 +312,16 @@ if (typeof browser === "undefined") {
        *        yield a response. False otherwise.
        */
       return function onMessage(message, sender, sendResponse) {
-        let result = listener(message, sender);
+        let didCallSendResponse = false;
+        let result = listener(message, sender, function(result) {
+          // Note: No need to check for duplicate calls. The browser environment
+          // should take care of reporting errors if necessary.
+          didCallSendResponse = true;
+          sendResponse(result);
+        });
+        if (didCallSendResponse || result === true) {
+          return result;
+        }
 
         if (isThenable(result)) {
           result.then(sendResponse, error => {
--- a/test/test-runtime-onMessage.js
+++ b/test/test-runtime-onMessage.js
@@ -155,9 +155,10 @@ describe("browser-polyfill", () => {
         wrappedListener("fake message", {name: "fake sender"}, sendResponseSpy);
 
         ok(messageListener.calledOnce, "The unwrapped message listener has been called");
-        deepEqual(messageListener.firstCall.args,
-                  ["fake message", {name: "fake sender"}],
-                  "The unwrapped message listener has received the expected parameters");
+        equal(messageListener.firstCall.args.length, 3, "expected argument count");
+        equal(messageListener.firstCall.args[0], "fake message", "expected message");
+        deepEqual(messageListener.firstCall.args[1], {name: "fake sender"}, "expected sender");
+        equal(typeof messageListener.firstCall.args[2], "function", "expected parameter");
 
         ok(sendResponseSpy.calledOnce, "The sendResponse function has been called");
         equal(sendResponseSpy.firstCall.args[0], "fake reply",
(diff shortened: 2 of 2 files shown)


nklein:
I futzed around for a bit before finding a solution to this same problem.  The above solution to the problem works well for me, too.  So, I've just started using this branch.  I look forward to seeing it merged.

slhck:
I used to implement message passing like this:

```javascript
chrome.tabs.sendMessage(id, { event : "foo" }, {}, (response) => {
    // do something
});

// handler
chrome.runtime.onMessage.addListener(function(request, sender, sendResponse) {
    sendResponse("bar");
});
```

I applied this patch in `browser-polyfill.js` and rewrote my code to this:

```javascript
browser.tabs.sendMessage(id, { event : "foo" }, {}).then((response) => {
    // do something
});

// handler
browser.runtime.onMessage.addListener(function(request, sender, sendResponse) {
    sendResponse(true);
    return Promise.resolve("bar");
});
```

However, the result of `browser.tabs.sendMessage` is always `undefined`. As far as I can tell, the code in the polyfill is executed normally, with the `Promise` being correctly created.

screenshot

Any ideas what could have gone wrong here?

slhck:
@nklein How did you implement this? Could you perhaps post a sample so I can double-check what's going wrong with my

nklein:
As I understand the documentation on the Mozilla Developer Network site, the listener is supposed to return 'true' if it has called or will call `sendResponse()`.  So, I would have expected your listener code to be:

```
sendResponse("bar");
return true;
```

Or, if it needed to be asynchronous, then:
```
var promise = new Promise( (resolve, reject) => {
    var result = do_something_long();
    resolve( "bar" );
} );
return true;
```

Or if you are doing something explicitly asynchronous, then:
```
doSomethingThatReturnsPromise.then( sendResponse, (err) => {
   sendResponse( { 'err': err } );
});
return true;
```

slhck:
Thanks! Actually, using the patch from @Rob--W – without changing his code – I just do this:

```javascript
return Promise.resolve("bar");
```

Which gives me the correct response, but a few seconds (like 10–20?) later, it shows errors in my console:

image

When I instead do:

```javascript
// background page
let ret = browser.tabs.sendMessage(id, "foo");
ret.then((response) => {
  // handle response
}, (error) => {
  // handle error
}).catch((error) => {
  // handle other errors
});

// content script
browser.runtime.onMessage.addListener((request, sender, sendResponse) => {
  sendResponse("bar");
  return(true);
}
```

I get an error inside `then()`: `The message port closed before a reponse was received`.

rpl:
Hi @Rob--W,
As I mentioned in #16 (https://github.com/mozilla/webextension-polyfill/issues/16#issuecomment-296693219), I discussed about this issue and the related pull requests with @kmaglione, and it looks like the `sendResponse` callback has not been included in the polyfill implementation on purpose because is going to be removed from the W3C specs.

Based on the above discussion, we agreed to mark the above issue as *needs: docs* (to fix the ambiguity in our MDN docs) and *state: wontfix* (to state that the deprecated `sendResponse` callback will not be implemented).

PR #27, opened 2017-03-22
Enable publishing to NPM

collinsauve:
I'm publishing this to an internal NPM server (sinopia).  I had to make these changes to be able to publish it.

Any reason why this package is not published on official NPM?


rpl:
Hi @collinsauve,
As mentioned by @josephfrazier, in #17 we are starting to wrap the built file as an UMD module, and so the polyfill should now be able to work as a commonjs module (and there is also a new test case for it).

I'm closing this PR because we just merged #17 and it should already cover this (feel free to open a new issue and/or pull request if anything is missing, so that we can work on them as "follow-ups" of #17).

My apologies for the duplicated effort and the waiting time.

Issue #29, opened 2017-03-22
Export as module


collinsauve:
I would like to use this as a module in webpack.   Would it be possible to publish a version that exports `browser`?

I made an attempt on this PR and it did work when I consume it as a module, however the way the tests load the script don't work: https://github.com/mozilla/webextension-polyfill/pull/28

collinsauve:
Refactored the way I do the export here: https://github.com/mozilla/webextension-polyfill/pull/31

rpl:
Fixed by #17

PR #31, opened 2017-03-22
Export browser object

collinsauve:


collinsauve:
For issue https://github.com/mozilla/webextension-polyfill/issues/29

BigBlueHat:
Might be best to squash the commits to make review easier. Not sure if something like UMD is needed in this case, but it would be very nice to `require('webextension-polyfill')` in the future.

Great idea @collinsauve

collinsauve:
Thanks @BigBlueHat, squashed.

josephfrazier:
https://github.com/mozilla/webextension-polyfill/pull/17 also has some changes that make it possible to `npm install` from github and use with a bundler. If it's merged and subsequently published to npm, it will also be possible to `npm install` the pre-built `dist/` files, so that no build tools or bundlers are needed.

---

If you'd like, you can go ahead and install it from github using:

    npm install --save https://github.com/mozilla/webextension-polyfill#99faff94ffbf03094361875a10e1692acc4b8662

rpl:
Hi @collinsauve,
As mentioned by @josephfrazier, in #17 we are starting to wrap the built file as an UMD module, and so the polyfill should now be able to work as a commonjs module (and there is also a new test case for it).

I'm closing this PR because we just merged #17 and it should already cover this (feel free to open a new issue and/or pull request if anything is missing, so that we can work on them as "follow-ups" of #17).

My apologies for the duplicated effort and the waiting time.

PR #40, opened 2017-06-16
Fix spacing to allow npm install to pass

mxie:
`npm install` was failing because of bad indentation. These changes address that.


Rob--W:
I'm closing this PR because #42 addresses this issue in a more robust way.

PR #43, opened 2017-07-13
Handle the browsers which don't support promises (eg. Edge)

kai918:
Compatibility with environments that have the browser namespace but do not support promises.

I was inspired by Snoak's answer in the following issue: mozilla#3

### DIFF
--- a/src/browser-polyfill.js
+++ b/src/browser-polyfill.js
@@ -14,7 +14,7 @@ try {
 }
 
 if (typeof browser === "undefined" || !_supportsPromises) {
-  var _browser = window.browser || window.msBrowser || window.chrome;
+  var _browser = window.browser || window.msBrowser || window.chrome || browser;
 
   // Wrapping the bulk of this polyfill in a one-time-use function is a minor
   // optimization for Firefox. Since Spidermonkey does not fully parse the
--- a/src/browser-polyfill.js
+++ b/src/browser-polyfill.js
@@ -6,7 +6,15 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 "use strict";
 
-if (typeof browser === "undefined") {
+var _supportsPromises = false;
+try {
+  _supportsPromises = browser.runtime.getPlatformInfo() instanceof Promise;
+} catch (e) {
+}
+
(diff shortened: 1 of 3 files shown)


Croydon:
What's up with this pull request?

mi-g:
This change is dedicated to handling browsers (Edge) that offer a `browser` global but do not provide a native promise return in WebExtensions methods.
The detection of this case is done by calling a method that requires a callback (necessary because no promise returns). If the method generates an exception, it means it does not support promise and the `webextension-polyfill` code must do its job.
Unfortunately in this PR, the method chosen to test the promise availability is `browser.runtime.getPlatformInfo()` which indeed generates an exception on Edge, but primarily because this method is not implemented in this browser ([see MDN documentation](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/runtime/getPlatformInfo)).
Using another method, like `browser.runtime.getBackgroundPage()` to test the promise availability would be a better idea.

src/browser-polyfill.js
```diff
@@ -6,7 +6,16 @@
...
 
-if (typeof browser === "undefined") {
+var _supportsPromises = false;
+try {
+  _supportsPromises = browser.runtime.getPlatformInfo() instanceof Promise;
+} catch (e) {
+  console.log("promises not supported.");
+}
+
+if (typeof browser === "undefined" || !_supportsPromises) {
```
ExE-Boss: A better option would be to change this line to:
```javascript
if (typeof browser === "undefined" || !(() => {
  let supportsPromises = false;
  try {
    supportsPromises = browser.runtime.getPlatformInfo() instanceof Promise;
  } catch (e) {
    console.log("promises not supported.");
  }
  return supportsPromises;
})()) {
```
And remove the previous block.

rpl:
Closing in favor of #114 (which is still exploring how we may allow the polyfill to work on Edge, and it is also taking advantage of the recently added integration tests to ensure that the additional changes needed are not going to break the polyfill behaviors expected on Chrome and Firefox).

lfjr:
So do I need to do anything to fix it other than the tools that I'm downloading

Issue #41, opened 2017-07-11
npm install fails with eslint error message


groovecoder:
On a fresh clone ...
```
groovecoder:webextension-polyfill lcrouch$ npm i
npm WARN prefer global coffee-script@1.10.0 should be installed with -g

> webextension-polyfill@0.1.1 prepublish /Users/lcrouch/code/webextension-polyfill
> npm run build && npm run test


> webextension-polyfill@0.1.1 build /Users/lcrouch/code/webextension-polyfill
> grunt

Running "eslint:src" (eslint) task

/Users/lcrouch/code/webextension-polyfill/Gruntfile.js
  36:1  error  Expected indentation of 18 spaces but found 28  indent
  37:1  error  Expected indentation of 18 spaces but found 28  indent

✖ 2 problems (2 errors, 0 warnings)
  2 errors, 0 warnings potentially fixable with the `--fix` option.

Warning: Task "eslint:src" failed. Use --force to continue.

Aborted due to warnings.

npm ERR! Darwin 15.6.0
npm ERR! argv "/Users/lcrouch/.nvm/versions/node/v7.0.0/bin/node" "/Users/lcrouch/.nvm/versions/node/v7.0.0/bin/npm" "run" "build"
npm ERR! node v7.0.0
npm ERR! npm  v3.10.8
npm ERR! code ELIFECYCLE
npm ERR! webextension-polyfill@0.1.1 build: `grunt`
npm ERR! Exit status 3
npm ERR!
npm ERR! Failed at the webextension-polyfill@0.1.1 build script 'grunt'.
npm ERR! Make sure you have the latest version of node.js and npm installed.
npm ERR! If you do, this is most likely a problem with the webextension-polyfill package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     grunt
npm ERR! You can get information on how to open an issue for this project with:
npm ERR!     npm bugs webextension-polyfill
npm ERR! Or if that isn't available, you can get their info via:
npm ERR!     npm owner ls webextension-polyfill
npm ERR! There is likely additional logging output above.

npm ERR! Please include the following file with any support request:
npm ERR!     /Users/lcrouch/code/webextension-polyfill/npm-debug.log

npm ERR! Darwin 15.6.0
npm ERR! argv "/Users/lcrouch/.nvm/versions/node/v7.0.0/bin/node" "/Users/lcrouch/.nvm/versions/node/v7.0.0/bin/npm" "i"
[...]

PR #48, opened 2017-07-19
Save 60% or more by switching to arrays today!

bfred-it:
I saw that the whole JSON file repeats the same keys so I thought of shortening it to arrays. This halves the size of the file and browser memory usage. JSDoc still describes the two indexes so IDEs don't get confused.

JSDoc format found here: https://github.com/jsdoc3/jsdoc/issues/1073#issuecomment-167204327

Failing build is due to #41

### DIFF
--- a/api-metadata.json
+++ b/api-metadata.json
@@ -1,512 +1,164 @@
 {
   "alarms": {
-    "clear": {
-      "minArgs": 0,
-      "maxArgs": 1
-    },
-    "clearAll": {
-      "minArgs": 0,
-      "maxArgs": 0
-    },
-    "get": {
-      "minArgs": 0,
-      "maxArgs": 1
-    },
-    "getAll": {
-      "minArgs": 0,
-      "maxArgs": 0
-    }
+    "clear": [0, 1],
+    "clearAll": [0, 0],
+    "get": [0, 1],
+    "getAll": [0, 0]
   },
   "bookmarks": {
-    "create": {
-      "minArgs": 1,
-      "maxArgs": 1
-    },
-    "export": {
-      "minArgs": 0,
-      "maxArgs": 0
-    },
-    "get": {
-      "minArgs": 1,
-      "maxArgs": 1
-    },
-    "getChildren": {
-      "minArgs": 1,
-      "maxArgs": 1
-    },
-    "getRecent": {
-      "minArgs": 1,
-      "maxArgs": 1
-    },
-    "getTree": {
-      "minArgs": 0,
-      "maxArgs": 0
-    },
-    "getSubTree": {
-      "minArgs": 1,
-      "maxArgs": 1
-    },
-    "import": {
-      "minArgs": 0,
-      "maxArgs": 0
-    },
-    "move": {
-      "minArgs": 2,
-      "maxArgs": 2
-    },
-    "remove": {
-      "minArgs": 1,
-      "maxArgs": 1
-    },
-    "removeTree": {
-      "minArgs": 1,
-      "maxArgs": 1
-    },
-    "search": {
-      "minArgs": 1,
-      "maxArgs": 1
-    },
-    "update": {
-      "minArgs": 2,
-      "maxArgs": 2
-    }
+    "create": [1, 1],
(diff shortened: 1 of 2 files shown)


bfred-it:
If there's interest I can resolve the conflicts.

bfred-it:
Actually there’s an even more efficient solution to this by using defaults for the most common situation. If PRs are considered I’ll replace this one.

ahmadassaf:
I like this PR but i think now with the new devtools support this will not work as the api metadata contains more info

```js
        "devtools": {
          "inspectedWindow": {
            "eval": {
              "minArgs": 1,
              "maxArgs": 2
            }
          },
          "panels": {
            "create": {
              "minArgs": 3,
              "maxArgs": 3,
              "singleCallbackArg": true
            }
          }
        },
```

bfred-it:
That’s ok, it could be minified to [3, 3, true] (and then whatever comes later). The solution I was thinking about was to automatically apply these optimizations in the gruntfile, so the repo can have the human-readable json and the outputted file can be small.

ExE-Boss:
Or, make the third argument an optional options object, similarly to what ESLint does.

james-cnz:
For what it's worth, I looked at how much could be saved on the JSON file (as it is at the moment) with various partial measures.

Methods are currently described in the JSON file like follows:
  "methodName": {
    "minArgs": n,
    "maxArgs": n,
    "fallbackToNoCallback": true
  },

Removing line breaks and indents from within the method description, like follows, saves ~26%
  "methodName": {"minArgs": n, "maxArgs": n, "fallbackToNoCallback": true},

Removing quote marks from within the method description, like follows, further saves ~6%
  "methodName": {minArgs: n, maxArgs: n, fallbackToNoCallback: true},

Using an array for the number of arguments, like follows, further saves ~14%
  "methodName": {args: [n, n], fallbackToNoCallback: true},

Using an array for the whole method description, like follows (and as per the PR), further saves ~13%
  "methodName": [n, n, true],

This adds up to a total saving of ~59%.

ExE-Boss:
> Removing quote marks from within the method description

That’s no longer valid JSON.

Rob--W:
>> Removing quote marks from within the method description
>
> That’s no longer valid JSON.

That is not a problem, because the JSON data is embedded as a JavaScript object literal in the polyfill.

bfred-it:
These optimizations can be done at bundling time. @Rob--W is it worth looking into it or should I abandon this PR?

rpl:
@bfred-it we were still changing the metadata for fixing issues specific to some of the API methods (and also add more of the API methods and namespaces) and so I preferred to defer big changes to the api-metadata json file format, nevertheless I still think that reducing the size of the JSON api metadata could be a very good thing.

If we can leave the api-metadata.json file in a human-readable format in the source repo, then this sounds even better to me, the optimizations we need to "make these JSON data smaller in memory" can be definitely be part of the build process.

james-cnz:
I didn't know the quotes are required in JSON.  I guess this needn't be problem, as Rob--W says, if the "JSON" file is considered to be a fragment of JavaScript, rather than JSON.  It might be nice to keep it as JSON, anyway, to allow easier automatic validation, though?

Personally, I don't like the idea of adding custom optimisations into the build process, because this would add a little bit more complexity, and I struggle to understand the polyfill as it is.

With regards to documenting arrays with differently-typed elements in JSDoc, further down in the comment linked from the original post, I think it says this is a "work around",  not official JSDoc?  So support for validating this might be variable?

It's also occurred to me, if the JavaScript is to be patched during the build process, to match the changes to the JSON, this could be fragile?  Whereas if the changes are made to the JavaScript source, this would reduce readability.

I think removing line breaks and the associated indents from within the method descriptions, together with using an array for the number of arguments, would save ~40%, without breaking the JSON or requiring any change to the build process?

Alternatively, looking at how much could be saved by removing comments from browser-polyfill.js, I think this works out as equivalent to ~96% of the size of the JSON file?  Could doing this in the build process be an easy win?

ExE-Boss:
I made PR #141 to always minify `api-metadata.json` when importing it.

PR #50, opened 2017-08-03
Drop reference to babel for async/await

bfred-it:
async/await is supported across the board now, no need for transpilers

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,9 +95,8 @@ The major differences are:
 * Rather than nesting callbacks when a sequence of operations depend on each
   other, Promise chaining is generally used instead.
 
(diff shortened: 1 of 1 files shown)


Rob--W:
Firefox is not relevant here, because the polyfill is no-op in Firefox.

FYI, Chrome supports async/await as of version 55 (the current stable version is 65).

ExE-Boss:
But Firefox is relevant to `async` and `await`, since async functions are used outside the polyfill code, and this PR is about removing the note about the requirement to use Babel to be able to use `async` and `await` in your own code.

Rob--W:
Oh I overlooked that this is a PR. The proposed change looks good to me too.

PR #51, opened 2017-08-12
Fix Constructor WeakMap requires 'new'

### LINKED ISSUE #14: ESLint causes build on windows to fail with default git settings

Zhouzi:
When using this polyfill in Chrome, I'm getting this error:

> Constructor WeakMap requires 'new'

After some google searches, it seems that extending built-in is not supported. This PR adds a babel plugin allowing that.

Also fix #14 

EDIT: Chrome version is 60.0.3112.90, OS is macOS Sierra 10.12.5

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -63,6 +63,11 @@ module.exports = function(grunt) {
           babelrc: false,
           comments: false,
           presets: ["babili"],
+          plugins: [
+            ["babel-plugin-transform-builtin-extend", {
+              globals: ["WeakMap"],
+            }],
+          ],
           sourceMap: true,
         },
         files: {
@@ -79,6 +84,9 @@ module.exports = function(grunt) {
                 "webextension-polyfill": "browser",
               },
               exactGlobals: true,
+            }],
+            ["babel-plugin-transform-builtin-extend", {
+              globals: ["WeakMap"],
             }],
           ],
           sourceMap: true,
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
   },
   "homepage": "https://github.com/mozilla/webextension-polyfill",
   "devDependencies": {
+    "babel-plugin-transform-builtin-extend": "^1.1.2",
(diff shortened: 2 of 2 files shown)


Gruntfile.js
```diff
@@ -33,8 +33,8 @@ module.exports = function(grunt) {
               match: /\{\/\* include\("(.*?)"\) \*\/\}/,
               replacement: (match, filename) => {
                 return grunt.file.read(filename)
-                            .replace(/\n$/, "")
-                            .replace(/^[^{]/gm, "    $&");
+                  .replace(/\n$/, "")
+                  .replace(/^[^{]/gm, "    $&");
```
Rob--W: Could you undo the change to these two lines?

Zhouzi:
@Rob--W I just reverted the changes to the mentioned lines.

Rob--W:
Could you squash all commits too? Otherwise `git blame` will still unnecessarily point to these commits.

Zhouzi:
Here you go, all in one 😄

Rob--W:
Wait, the build script of webextension-polyfill itself does not transpile the source to ES5, so the `class DefaultWeakMap extends WeakMap` syntax should work without any issues in Chrome.

If we used a transpiler, then this patch would be good to go. However, we don't transpile, so why should this PR be merged?

Zhouzi:
I double checked and in fact it works as expected as long as the code is not transpiled. The error comes from the transpiled version of the code (using babel). Which means the issue is not much of webextension-polyfill's business. Sorry for the mistake!

PR #54, opened 2017-08-28
Make extensions more testable by allowing a custom 'chrome' object

nspaeth:


Rob--W:
Thanks for opening this PR. However, I don't think that this is necessary, since it is already possible to test the API, e.g. by stubbing or spying on the global `chrome` object in a "before" hook (and restore the original state in an "after" hook) (many test frameworks offer the ability to set up the environment before running the actual tests).

nspaeth:
Yes, this can probably be closed.

While trying to get testing working I was getting errors for 'chrome' being undefined so I added this line. Eventually it started working but I think mistakenly attributed it to this change because it is still working for me without this change.

PR #57, opened 2017-09-23
feat: Add support for wrapping the devtools APIs

### LINKED ISSUE #56: browser.devtools.* isn't polyfilled
First of all, thanks for writing and maintaining this polyfill! 

However...I'm trying to port a Chrome extension to Firefox and having a difficult time, as it seems that not all API methods are polyfilled. One example is [browser.devtools.panels.create](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/devtools.panels/create), which uses a callback signature in Chrome but is promise-based in Firefox. I've got the compiled polyfill implemented and Chrome is throwing an error:

```
Uncaught TypeError: Cannot read property 'then' of undefined
    at Object.2../actions/inspect (devtools.js:39)
    at s (devtools.js:1)
    at e (devtools.js:1)
    at devtools.js:1
```
That line points to this code:
```
browser.devtools.panels.create(name, '', 'panel/index.html').then((extensionPanel)  => {
//....
})
```
Is there a list of all the polyfilled methods somewhere, and/or perhaps a status for when the missing methods will be implemented? 

Edit: I see the metadata file does not include any devtools APIs. Bummer.

rpl:
This PR is applying the following changes to fix #56:

- the changes into api-metadata.json, needed to instruct the polyfill to wrap the devtools API namespace)
- a fix in the base browser-polyfill.js module to prevent the `TypeError` exception when the getter on the proxy is related to a non-configurable read-only property of the original Proxy target (fixed by using an empty object which has the original target as its prototype as the "root" target of the Proxy instance created by the polyfill)
- a fix in the base browser-polyfill.js module to force devtools.panels.create to return a promise that resolves to a single callback argument (instead of an resolving to an array of arguments which contains an additional undocumented second parameter)
- fix applied on one of the existent tests (related to the `Proxy TypeError` fix)
- 2 additional test case (which covers the `Proxy TypeError` fix and the new `maxResolvedArgs` property in the `devtools.panels.create` metadata).

### DIFF
--- a/src/browser-polyfill.js
+++ b/src/browser-polyfill.js
@@ -76,15 +76,20 @@ if (typeof browser === "undefined") {
      *        The promise's resolution function.
      * @param {function} promise.rejection
      *        The promise's rejection function.
+     * @param {object} metadata
+     *        Metadata about the wrapped method which has created the callback.
+     * @param {integer} metadata.maxResolvedArgs
+     *        The maximum number of arguments which may be passed to the
+     *        callback created by the wrapped async function.
      *
      * @returns {function}
      *        The generated callback function.
      */
-    const makeCallback = promise => {
+    const makeCallback = (promise, metadata) => {
       return (...callbackArgs) => {
         if (chrome.runtime.lastError) {
           promise.reject(chrome.runtime.lastError);
-        } else if (callbackArgs.length === 1) {
+        } else if (metadata.singleCallbackArg || callbackArgs.length === 1) {
           promise.resolve(callbackArgs[0]);
         } else {
           promise.resolve(callbackArgs);
@@ -107,6 +112,9 @@ if (typeof browser === "undefined") {
      *        The maximum number of arguments which may be passed to the
      *        function. If called with more than this number of arguments, the
      *        wrapper will raise an exception.
+     * @param {integer} metadata.maxResolvedArgs
+     *        The maximum number of arguments which may be passed to the
+     *        callback created by the wrapped async function.
      *
      * @returns {function(object, ...*)}
      *       The generated wrapper function.
@@ -124,7 +132,7 @@ if (typeof browser === "undefined") {
         }
 
         return new Promise((resolve, reject) => {
-          target[name](...args, makeCallback({resolve, reject}));
+          target[name](...args, makeCallback({resolve, reject}, metadata));
         });
       };
     };
@@ -340,7 +348,12 @@ if (typeof browser === "undefined") {
       },
     };
 
-    return wrapObject(chrome, staticWrappers, apiMetadata);
+    // Create an object that has the real target as its prototype
+    // to prevent a Proxy violation exception for the devtools API getter
+    // (which is a read-only non-configurable property on the original target).
+    const targetObject = Object.create(chrome);
+
+    return wrapObject(targetObject, staticWrappers, apiMetadata);
   };
 
   // The build process adds a UMD wrapper around this file, which makes the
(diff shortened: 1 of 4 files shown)


rpl:
Hi @Rob--W 
do you mind to take a look at this PR?

src/browser-polyfill.js
```diff
@@ -76,18 +76,30 @@ if (typeof browser === "undefined") {
...
         if (chrome.runtime.lastError) {
           promise.reject(chrome.runtime.lastError);
-        } else if (callbackArgs.length === 1) {
-          promise.resolve(callbackArgs[0]);
         } else {
-          promise.resolve(callbackArgs);
+          if ("maxResolvedArgs" in metadata) {
+            // Some chrome APIs provides to their callback more arguments than expected.
+            callbackArgs = callbackArgs.slice(0, metadata.maxResolvedArgs);
+          }
```
Rob--W: There is no need to be so generic. Since the problem only occurs when with 1 vs not-one parameter, I suggest to throw away "maxResolvedArgs" and use `"singleCallbackParameter": true` or something similar. This could also be used to fix API definitions for callbacks that don't expect any parameter. Based on the code below, it is likely that callbacks without parameters resolve to an empty array.

test/test-async-functions.js
```diff
@@ -86,5 +86,30 @@ describe("browser-polyfill", () => {
...
     });
+
+    it("resolves the expected number of arguments on metadata maxResolvedArgs property", () => {
+      const fakeChrome = {
+        runtime: {lastError: null},
+        devtools: {
+          panels: {
+            create: sinon.spy((title, iconPath, panelURL, cb) => {
+              Promise.resolve().then(() => {
+                cb({isFakePanel: true}, "unwanted parameter");
```
Rob--W: Leave a comment what happens with `"unwanted parameter"`, and why.

test/test-proxied-properties.js
```diff
@@ -75,28 +105,28 @@ describe("browser-polyfill", () => {
     });
 
     it("deletes proxy getter/setter that are not wrapped", () => {
-      const fakeChrome = {};
+      const fakeChrome = {runtime: {}};
```
Rob--W: Why these changes? Please explain the changes in the commit description.

Rob--W:
I have filed an upstream bug for Chrome's misbehaving devtools.panels.create API: https://bugs.chromium.org/p/chromium/issues/detail?id=768159

rpl:
@Rob--W Thanks a lot for the review comments (and for filing the issue on Chromium issue tracker), I've applied the following changes based on you review comments:

- changed the new metadata property to `singleCallbackArg`, and added a test case related to a callback which expects 1 parameter called without any parameter (so that we test explicitly that the promise will resolve to a single undefined parameter instead of an empty array of callback parameters), and added to the test cases an inline comment that explains what behavior we expect

- squashed all the commits upfront and added a commit description comment which explains the changes applied (which included some additional information about the reason of the changes applied to the existent test case related to the "proxy setter/getter behaviors on non wrapped properties"): https://github.com/mozilla/webextension-polyfill/pull/57/commits/676966d0c28f5671d354de2c2882009db27e3ad4
tests

marcysutton:
Thank you so much for this! Excited to give it a go. Will report back in a new issue if I run into any problems.

marcysutton:
Do you know when this will get released? I'm not seeing it in 0.1.2. Thanks!

rpl:
@marcysutton I'm going to release a new version today (I'd like to include also #59, but if we are not able to land it today yet, I will release a new version for this change and I will include #59 in the next one).

Issue #24, opened 2017-02-16
Chrome: Error in event handler for tabs.onUpdated: Error: Expected at most 0 arguments for show(), got 1


T-vK:
![Screenshot](https://i.snag.gy/CaOIpV.jpg)

My extension works without any errors in the latest version of Firefox. In Chrome I get this error.
Honestly, I'm not sure what is causing this error. I am using `tabs.onUpdated` and within that I'm using `pageAction.show()`. If you need more information, tell me.

Edit:
Maybe the issue is that I'm calling browser.pageAction.show() with a tab id in as the first argument as required by Firefox. I guess Chrome doesn't support it? 
I tried to work around it using `browser.pageAction.show(browser.pageAction.show.length?tabId:undefined)`.   
But that didn't make any difference.

Edit2: 
It seems to be a misconfiguration in the browser-polifill library. Setting `minArgs` and `maxArgs` of `show` under pageAction to 1 seems to fix it.

Edit3:
Actually I'm still getting an error:  
`Error: Invocation of form pageAction.show(integer, function) doesn't match definition pageAction.show(integer tabId)`

Edit4:
Here is the code causing the error:
``` JavaScript
browser.tabs.query({}).then( tabs => {
    if (!tabs)
        return
    let promiseArray = []
    tabs.forEach(tab => {
        let tabId = tab.id
        promiseArray.push(browser.pageAction.show(tabId).then(()=>{
            return browser.pageAction.setIcon({tabId: tabId, path: 'disabled.ico'})
        }))
    })
    return Promise.all(promiseArray)
}).then(()=>{
    /* ... */
}).catch(console.error)
```

Rob--W:
See #13 (and #9). For now, just use chrome.pageAction.setIcon (or apply PR #13).

Rob--W:
Fixed by #59.

PR #59, opened 2017-09-26
Fix: Support pageAction show/hide wrapping

rpl:
This PR contains the changes to the api-metadata.json from #9 and it also applies the changes needed into the polyfill wrappers to be able to customize the wrapper behavior for the API methods that returns a Promise on Firefox but do not accept a callback on Chrome (which is the case for pageAction.show and hide).

### DIFF
--- a/src/browser-polyfill.js
+++ b/src/browser-polyfill.js
@@ -132,7 +132,39 @@ if (typeof browser === "undefined") {
         }
 
         return new Promise((resolve, reject) => {
-          target[name](...args, makeCallback({resolve, reject}, metadata));
+          function callAPIWithNoCallback() {
+            try {
+              target[name](...args);
+
+              // Update the API method metadata, so that the next API calls will not try to
+              // use the unsupported callback anymore.
+              metadata.fallbackToNoCallback = false;
+              metadata.noCallback = true;
+
+              resolve();
+            } catch (error) {
+              // Catch API parameters validation errors and rejects the promise.
+              reject(error);
+            }
+          }
+
+          if (metadata.fallbackToNoCallback) {
+            // This API method has currently no callback on Chrome, but it return a promise on Firefox,
+            // and so the polyfill will try to call it with a callback first, and it will fallback
+            // to not passing the callback if the first call fails.
+            try {
+              target[name](...args, makeCallback({resolve, reject}, metadata));
+            } catch (cbError) {
+              console.warn(`${name} API method doesn't seem to support the callback parameter, ` +
+                           "falling back to call it without a callback: ", cbError);
+
+              callAPIWithNoCallback();
+            }
+          } else if (metadata.noCallback) {
+            callAPIWithNoCallback();
+          } else {
+            target[name](...args, makeCallback({resolve, reject}, metadata));
+          }
         });
       };
     };
--- a/src/browser-polyfill.js
+++ b/src/browser-polyfill.js
@@ -132,22 +132,6 @@ if (typeof browser === "undefined") {
         }
 
         return new Promise((resolve, reject) => {
-          function callAPIWithNoCallback() {
-            try {
-              target[name](...args);
-
-              // Update the API method metadata, so that the next API calls will not try to
-              // use the unsupported callback anymore.
-              metadata.fallbackToNoCallback = false;
-              metadata.noCallback = true;
-
-              resolve();
-            } catch (error) {
-              // Catch API parameters validation errors and rejects the promise.
-              reject(error);
-            }
-          }
-
           if (metadata.fallbackToNoCallback) {
             // This API method has currently no callback on Chrome, but it return a promise on Firefox,
             // and so the polyfill will try to call it with a callback first, and it will fallback
@@ -158,10 +142,18 @@ if (typeof browser === "undefined") {
               console.warn(`${name} API method doesn't seem to support the callback parameter, ` +
                            "falling back to call it without a callback: ", cbError);
 
-              callAPIWithNoCallback();
+              target[name](...args);
+
+              // Update the API method metadata, so that the next API calls will not try to
+              // use the unsupported callback anymore.
+              metadata.fallbackToNoCallback = false;
+              metadata.noCallback = true;
+
+              resolve();
             }
           } else if (metadata.noCallback) {
-            callAPIWithNoCallback();
+            target[name](...args);
+            resolve();
           } else {
             target[name](...args, makeCallback({resolve, reject}, metadata));
           }
--- a/api-metadata.json
+++ b/api-metadata.json
@@ -300,16 +300,16 @@
       "maxArgs": 1
     },
     "hide": {
-      "minArgs": 0,
-      "maxArgs": 0
+      "minArgs": 1,
+      "maxArgs": 1
     },
     "setIcon": {
       "minArgs": 1,
       "maxArgs": 1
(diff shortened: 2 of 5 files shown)


rpl:
@Rob--W do you mind to take a look at this PR?

similarly to #57 it is going to change the api-metadata.json but it also needs some additional changes in the polyfill wrappers, to be able to customize the behavior for the kind on chrome incompatibility that pageAction.show/hide presents.

Rob--W:
This doesn't look like a forward-compatible change to me.

Yes, it would fix the current error.
However, if Chrome were to support the callback parameter (which I expect to happen - https://github.com/mozilla/webextension-polyfill/pull/13#issuecomment-269387289, I've wanted to submit a patch myself, but that's near the bottom of my priority list), then the polyfill will cause an error again.

How about trying to call with callback, and if a synchronous (schema) error about unsupported callbacks occurs, invoke the method without callback?

```javascript
try { chrome.browserAction.setTitle(1, function(){}); } catch (e) { console.log(e.message); }
```

gives the following result (tested Chrome 34 - 61:

> Invocation of form browserAction.setTitle(integer, function) doesn't match definition browserAction.setTitle(object details)

and when `--native-crx-bindings=1` is set (tested with Chrome 61) (Chrome is working on [replacing JS-implemented extension bindings with C++](https://bugs.chromium.org/p/chromium/issues/detail?id=653596)):

> Error in invocation of browserAction.setTitle(object details): Too many arguments.


PS. Why does browserAction.setTitle not appear in api-metadata.json?

src/browser-polyfill.js
```diff
@@ -153,6 +150,27 @@ if (typeof browser === "undefined") {
...
+            // This API method has currently no callback on Chrome, but it return a promise on Firefox,
+            // and so the polyfill try to pass a callback but fallback to not passing the callback
+            // if the first call have been rejected.
+            try {
+              target[name](...args, makeCallback({resolve, reject}, metadata));
+            } catch (cbError) {
+              console.warn(`${name} API method doesn't seem to support the callback parameter, ` +
+                           "falling back to call it without a callback: ", cbError);
+
+              // Update the API method metadata, so that the next API call will not try to
```
Rob--W: Move this after line 170, to make sure that the API call is actually valid before considering the API callback-less.

Rob--W:
The `browserAction` API map is missing entries like `browser.runtime.enable`.

I guess that the JSON file is manually edited. Maybe it helps if you write a quick script to load the JSON files from mozilla-central and compare it with the existing definitions to see if anything is missing.

src/browser-polyfill.js
rpl: :+1: great call!

rpl:
@Rob--W I've updated this PR based on your last round of review comments and updated the test cases accordingly, let me know how it looks to you.

Thanks again for you help!

src/browser-polyfill.js
```diff
@@ -132,7 +132,48 @@ if (typeof browser === "undefined") {
         }
 
         return new Promise((resolve, reject) => {
-          target[name](...args, makeCallback({resolve, reject}, metadata));
+          function callAPIWithNoCallback() {
+            try {
+              target[name](...args);
+              if (chrome.runtime.lastError) {
```
Rob--W: This `chrome.runtime.lastError` check seems misplaced. The `target[name](...args)` call is asynchronous, so the error will never be set to that.

Worse, with the current code, if someone tries to use call e.g. `browser.pageAction.show` inside a function where `lastError` is set, then the promise will always be rejected.

src/browser-polyfill.js
```diff
@@ -132,7 +132,48 @@ if (typeof browser === "undefined") {
...
+          }
+
+          if (metadata.fallbackToNoCallback) {
+            // This API method has currently no callback on Chrome, but it return a promise on Firefox,
+            // and so the polyfill try to pass a callback but fallback to not passing the callback
+            // if the first call have been rejected.
+            try {
+              target[name](...args, makeCallback({resolve, reject}, metadata));
+            } catch (cbError) {
+              console.warn(`${name} API method doesn't seem to support the callback parameter, ` +
```
Rob--W: Under normal circumstances, this message will be displayed once.

This message will be displayed more than once if the extension passes the wrong arguments. Do you consider that to be OK? (personally I think that it is fine, since the code is simpler in this way; if you want to force the message to be displayed once, then it needs to be moved inside the `else` block above.

Why do we need log spam in the first place?

src/browser-polyfill.js
```diff
@@ -132,7 +132,48 @@ if (typeof browser === "undefined") {
         }
 
         return new Promise((resolve, reject) => {
-          target[name](...args, makeCallback({resolve, reject}, metadata));
+          function callAPIWithNoCallback() {
+            try {
+              target[name](...args);
+              if (chrome.runtime.lastError) {
+                // NOTE: Ideally this should rejects if there are runtime issues with
+                // the API call on Chrome, unfortunately currently when pageAction.show/hide
```
Rob--W: This whole comment block will go probably go away because of my above comment. But if you decide to re-instate the comments somewhere, generalize the affected APIs, e.g. "APIs such as pageAction.show" because pageAction.show/hide are not the only affected APIs.

src/browser-polyfill.js
rpl: :+1: it was already supposed to be removed and I forgot, thanks for pointing it out.

src/browser-polyfill.js
rpl: I thought to emit at least a warning log mostly because it is not completely impossible that this runtime discovery can be fooled by some chrome behaviors (maybe a future chrome behavior) that the polyfill is not expecting, and it could completely hide the real error.

Emitting the warning every time we are actually falling back between the "with callback" to the "no callback" mode is fine from my current point of view, in the common scenarios it should only be produced once, and if it is produced more than once, then it could be a signal of something different between the behavior that it expect and the actual one.

rpl:
@Rob--W suggested changes applied in the updated patch

src/browser-polyfill.js
```diff
@@ -132,7 +132,39 @@ if (typeof browser === "undefined") {
...
+              // use the unsupported callback anymore.
+              metadata.fallbackToNoCallback = false;
+              metadata.noCallback = true;
+
+              resolve();
+            } catch (error) {
+              // Catch API parameters validation errors and rejects the promise.
+              reject(error);
+            }
+          }
```
Rob--W: Now the `callAPIWithNoCallback` logic is very simple, it can be inlined below, without try-catch (because you are in the closure of a Promise).

The second `callAPIWithNoCallback` call does not require the metadata to be updated, so it can simply be:

```javascript
target[name](...args);
resolve();
```

Treora:
Would be great if this can be merged indeed, I stumbled upon #24 as well. Is this PR still held back by anything?

I notice one unprocessed review [comment](https://github.com/mozilla/webextension-polyfill/pull/59#discussion_r142001949); I made a [PR](https://github.com/rpl/webextension-polyfill/pull/1) to this branch that inlines the function as indicated.

rpl:
@Treora Thanks a lot for your PR! I've rebased this branch, added your changes on top of the commit already part of this PR and added pageAction.show to the smoke tests that runs on Chrome.

I'm going to land this PR as soon as Travis job has been completed to run.

PR #61, opened 2017-10-01
Made compatible with Edge browser

mi-g:
Detect the need to wrap the API even if global 'browser' object is defined (case of Edge) but methods do not support returning promises.
Adapt wrapping to Edge specificities.

### DIFF
--- a/src/browser-polyfill.js
+++ b/src/browser-polyfill.js
@@ -6,7 +6,15 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 "use strict";
 
-if (typeof browser === "undefined") {
+if ((() => {
+  try {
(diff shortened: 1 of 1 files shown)


mi-g:
This PR has 3 errors in Travis CI.
I believe the tests have to be reworked. For instance, the failed test "browser-polyfill does (not) override the global browser namespace if it already exists" is definitely wrong: Edge has a 'browser' global that has to be overridden. I'm unsure about the other 2 but the PR leaves the wrapping unchanged in case of Chrome browser.

ExE-Boss:
I’ve added tests to cover the specific issues in #114.

rpl:
Closing in favor of #114 (which is still exploring how we may allow the polyfill to work on Edge, and it is also taking advantage of the recently added integration tests to ensure that the additional changes needed are not going to break the polyfill behaviors expected on Chrome and Firefox).

PR #62, opened 2017-10-01
introducing table of contents

tiagomoraismorgado88:
# Greetings

**the current PR does basically aim at:**
- *introducing table of contents*

---

**the motivation behind such a thing is basically to:**
- *improve accessability of the file*
- *turning the file easier to be read*
- *turning the file more accessible to be read*


andymckay:
Thanks

PR #64, opened 2017-10-09
fix: Prevent 'Previous API instantiation failed' errors on runtime.sendMessage API calls

### LINKED ISSUE #58: runtime.sendMessage/onMessage
Hello

I'm running into problems when trying to use `browser.runtime.sendMessage` and `browser.runtime.onMessage` with this polyfill on Chrome 61.

I have this code that works fine:

```js
// in background script, run when the extension is loaded
chrome.runtime.onMessage.addListener((message, sender) => {
  console.log("background script got message", message, "from", sender);
});

// in content script, run when the user presses a button
chrome.runtime.sendMessage({type: "event", eventName: "THE_EVENT"});
```

If I replace `chrome` with `browser` in the background script, I get this in its console:

image

and the listener is *not* run.

Independently, if I replace `chrome` with `browser` on the content script side, I get this in the console of the page:

image

but the listener *is* run.

If I replace `chrome` with `browser` on both sides, then the content script says this instead:

image

and the background script message remains unchanged. 

I'm new to writing browser extensions so I'm not sure if this might be a bug in the polyfill or if not using the API quite right.

rpl:
This PR contains changes which aims to fix #58 using a strategy that it is also able to fix the `Proxy TypeError` issues related to the devtools API namespace (#57).

### DIFF
--- a/src/browser-polyfill.js
+++ b/src/browser-polyfill.js
@@ -348,10 +348,10 @@ if (typeof browser === "undefined") {
       },
     };
 
-    // Create an object that has the real target as its prototype
+    // Create a new empty object and copy the properties of the original chrome object
     // to prevent a Proxy violation exception for the devtools API getter
     // (which is a read-only non-configurable property on the original target).
-    const targetObject = Object.create(chrome);
+    const targetObject = Object.assign({}, chrome);
 
     return wrapObject(targetObject, staticWrappers, apiMetadata);
   };
--- a/test/test-proxied-properties.js
+++ b/test/test-proxied-properties.js
@@ -8,8 +8,10 @@ const {setupTestDOMWindow} = require("./setup");
 describe("browser-polyfill", () => {
   describe("proxies non-configurable read-only properties", () => {
     it("creates a proxy that doesn't raise a Proxy violation exception", () => {
-      const fakeChrome = {};
+      const fakeChrome = {"devtools": {}};
 
+      // Override the property to make it non-configurable (needed to be sure that
+      // the polyfill is correctly workarounding the Proxy TypeError).
       Object.defineProperty(fakeChrome, "devtools", {
         enumarable: true,
         configurable: false,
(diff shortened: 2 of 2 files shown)


rpl:
@Rob--W I took a look at the issues described in #58, and I've changed the strategy used to fix the devtools API namespace issues accordingly.

I've also tested this version manually on two small test extensions (one which reproduce the STR in #58 and one that uses the devtools API).

It would be nice to be able to include some integration tests which runs a test extension on a real Chromium instance (the devtools API could be a bit harder because the devtools page will only run if the user opens the developer tools on a tab, but testing the runtime.onMessage/sendMessage should be simpler), the unit tests are helpful to prevent regressions related to issues that we already know about, but they are not helpful to catch incompatibilities with the way Chromium build the `chrome` API object (and this kind of incompatibilities can also change between Chromium versions, e.g. once the new native bindings, which can be currently enabled using the `--native-crx-bindings=1` CLI options, becomes the default).

test/test-proxied-properties.js
```diff
@@ -8,8 +8,10 @@ const {setupTestDOMWindow} = require("./setup");
 describe("browser-polyfill", () => {
   describe("proxies non-configurable read-only properties", () => {
     it("creates a proxy that doesn't raise a Proxy violation exception", () => {
-      const fakeChrome = {};
+      const fakeChrome = {"devtools": {}};
 
+      // Override the property to make it non-configurable (needed to be sure that
+      // the polyfill is correctly workarounding the Proxy TypeError).
```
kumar303: 👍 for workarounding, I like it

src/browser-polyfill.js
```diff
@@ -348,10 +348,10 @@ if (typeof browser === "undefined") {
       },
     };
 
-    // Create an object that has the real target as its prototype
+    // Create a new empty object and copy the properties of the original chrome object
     // to prevent a Proxy violation exception for the devtools API getter
     // (which is a read-only non-configurable property on the original target).
-    const targetObject = Object.create(chrome);
+    const targetObject = Object.assign({}, chrome);
```
kumar303: Would it help to do `Object.freeze(Object.assign({}, chrome))` to make it read-only? I may not be understanding that part right though.

rpl: I'm not sure that Object.freeze the target object is really useful (it is also never exposed outside of the Proxy instance that uses it), I'm going to take a look at it once we have the smoke test (#66) in place.

rpl:
@kumar303 Thanks for the quick review.

PR #66, opened 2017-10-10
 test: Run a minimal set of tests from a test extension running on Chrome 

rpl:
This PR, as a follow up to #64, contains a prototype for a minimal integration test strategy:

Currently it runs a single test extension which includes the built polyfill on Chrome and uses the runtime.onMessage/sendMessage APIs to exchange a message between a background page and a content script (using puppeteer to control the Chrome browser instance, and the existent mocha and chai dev dependencies as the test runner and assertion library).

Using a similar strategy, we should also be able to run another smoke test for the devtools API as well (e.g. using the additional `--auto-open-devtools-for-tabs` command line parameter)


rpl:
Here is the link of the smoke test output when executed on travis: 
- https://travis-ci.org/mozilla/webextension-polyfill/builds/286046483#L1906-L1933

package.json
```diff
@@ -17,10 +17,14 @@
   },
   "homepage": "https://github.com/mozilla/webextension-polyfill",
   "devDependencies": {
+    "async-wait-until": "^1.1.5",
+    "babel-eslint": "^8.0.1",
     "babel-plugin-transform-es2015-modules-umd": "^6.24.1",
     "babel-preset-babili": "0.0.10",
```
kumar303: while you're in here you may want to add a caret to this dep

package.json
```diff
@@ -17,10 +17,14 @@
...
+    "babel-preset-es2017": "^6.24.1",
     "chai": "^3.5.0",
     "eslint": "3.9.1",
```
kumar303: this one is missing a caret too (not sure if that was intentional)

test/integration/test-runtime-messaging-on-chrome.js
```diff
@@ -0,0 +1,74 @@
...
+    ];
+
+    const lastExpectedMessage = expectedConsoleMessages.slice(-1).pop();
+
+    // Wait until the last expected message has been received.
+    const res = await waitUntil(() => {
+      return pageConsoleMessages.filter((msg) => {
+        return msg[0] === lastExpectedMessage[0] && msg[1] === lastExpectedMessage[1];
+      }).length > 0;
+    }, 5000).catch(error => error);
```
kumar303: what is this catch doing? Is it to prevent throwing an error so that you can throw it later? If so, a comment explaining that would be helpful.

test/integration/test-runtime-messaging-on-chrome.js
```diff
@@ -0,0 +1,74 @@
...
+
+    deepEqual(pageConsoleMessages, expectedConsoleMessages, "Got the expected console messages");
+
+    await browser.close();
+
+    server.close();
```
kumar303: should this `await server.close()` too?

test/integration/test-runtime-messaging-on-chrome.js
```diff
@@ -0,0 +1,74 @@
...
+
+    if (res instanceof Error) {
+      // Re-throw the unexpected error, if the test didn't fail yet.
+      throw res;
```
kumar303: What if you just wrapped it in a try/finally instead? Like:

```js
try {
  const res = await waitUntil(() => {
    // ...
  });
} finally {
  await browser.close();
  server.close();
}
```

test/run-chrome-smoketests.sh
```diff
@@ -0,0 +1,13 @@
+echo "\nTest webextension-polyfill from an extension running on chrome"
+echo "==============================================="
+
+export TEST_EXTENSIONS_PATH=/tmp/browser-polyfill-chrome-smoketests
+
+## Exits immediately if the directory already exists (which can only happen in a local
+## development environment, while this test will usually run on travis).
+mkdir $TEST_EXTENSIONS_PATH || exit 1
```
kumar303: I think it would be convenient for local testing if this script could be run repeatedly.

You could just remove the dir first which is probably pretty safe since it's a tmp dir:

```
rm -fr $TEST_EXTENSIONS_PATH
mkdir $TEST_EXTENSIONS_PATH || exit 1
```

Or if you wanted to be a bit paranoid you could check for a marker file to know if it's safe to delete:

```
MARKER_FILE=$TEST_EXTENSIONS_PATH/.created-for-run-chrome-smoketests
# Check if the marker file exists and then remove the directory
if [ ... $MARKER_FILE ]; then
  rm -fr $TEST_EXTENSIONS_PATH
fi
mkdir $TEST_EXTENSIONS_PATH || exit 1
touch $MARKER_FILE
```

And if you wanted to be *super* paranoid you could manually prefix `/tmp` for dangerous operations:

```
SUBDIR=browser-polyfill-chrome-smoketests
export TEST_EXTENSIONS_PATH=/tmp/$SUBDIR

MARKER_FILE=/tmp/$SUBDIR/.created-for-run-chrome-smoketests
# Check if the marker file exists and then remove the directory
if [ ... $MARKER_FILE ]; then
  rm -fr /tmp/$SUBDIR
fi
mkdir $TEST_EXTENSIONS_PATH || exit 1
touch $MARKER_FILE
```

kumar303:
Also, since I added some suggestions for cleanup, just let me know if you'd like a re-review.

test/run-chrome-smoketests.sh
rpl: :+1: I like it

I've included the "moderately paranoid" version in the updated version of this patches :-)

test/integration/test-runtime-messaging-on-chrome.js
rpl: :+1: I rewrote this patch to use a approach similar to the one suggested 

(slightly tweaked because I want to be sure to always include in the test failures both unexpected errors raised by waitUntil and differences between the expected console messages and the actual one that have been collected).

PR #67, opened 2017-10-12
Always use LF for dist files

yfdyh000:
This avoids got inconsistent line breaks and hash for `browser-polyfill.min.js` when build on Windows.

ref: https://github.com/gruntjs/grunt-contrib-concat/issues/20#issuecomment-15540197


Rob--W:
I wonder why this hasn't merged yet. Meanwhile I've submitted #107 with the `.gitattributes` change.

PR #71, opened 2017-10-23
Lazily initialize API via the original target

### LINKED ISSUE #70: clipboardWrite in manifest.json produces error in Chrome with v0.2.1 of the polyfill
I have [a browser extension](https://github.com/ipfs/ipfs-companion) that requires `clipboardWrite` permission under Firefox.
(copying under Chrome works without any additional permissions)

Everything works fine with webextension-polyfill v0.1.2 (the same code running in both Firefox and Chrome) but if I update webextension-polyfill  to v0.2.1 it fails to load in Chrome due to this error:
```
chrome.clipboard is not supported on this platform or manifest version
```

> chrome.clipboard is not supported on this platform or manifest version

If I remove `clipboardWrite` permission from `manifest.json` then it loads correctly in Chrome, but clipboard manipulation stops working in Firefox 🌵 

Looks like a bug, but maybe I missing something obvious here?
Let me know if I can test something or provide more debug information.

Rob--W:
@rpl Please take a look (see the commit message).

This fixes #70.

### DIFF
--- a/src/browser-polyfill.js
+++ b/src/browser-polyfill.js
@@ -191,13 +191,12 @@ if (typeof browser === "undefined") {
      */
     const wrapObject = (target, wrappers = {}, metadata = {}) => {
       let cache = Object.create(null);
-
       let handlers = {
-        has(target, prop) {
+        has(proxyTarget, prop) {
           return prop in target || prop in cache;
         },
 
-        get(target, prop, receiver) {
+        get(proxyTarget, prop, receiver) {
           if (prop in cache) {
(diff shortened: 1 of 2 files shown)


z543:
@Rob--W 
I got error "impl is not an instance of EventImpl"
using browser.runtime.onMessage.addListener

Rob--W:
@z543 I've fixed the problem and added more unit tests to cover the scenario.

z543:
It works now, thank you.

PR #83, opened 2018-01-07
Add metadata of sessions  api

kumabook:
Adding browser.sessions apis

### DIFF
--- a/api-metadata.json
+++ b/api-metadata.json
@@ -340,6 +340,48 @@
       "maxArgs": 1
     }
   },
+  "sessions": {
+    "forgetClosedTab": {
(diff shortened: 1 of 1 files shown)


kumabook:
Hi!
I found that there aren't sessions apis in ` api-metadata.json`.
However, chrome doesn't support `forgetClosedXXX` methods,
So I'm not sure that theses polyfill of session api should be included.

But I want to use these sessions api polyfill with my addons.

Thanks.

rpl:
Closing in favor of #126, which included the missing session API namespace (and some other missing API namespace) into the api metadata.

PR #84, opened 2018-01-07
[Docs] - Mozilla/webextension-polyfill - README.md - adding separators

tiagomoraismorgado88:
#### [Docs] - Mozilla/webextension-polyfill - README.md - adding separators


rpl:
The docs needs to be improved but the separators doesn't look to be needed.

PR #86, opened 2018-01-10
Improve browser detection fix #68

### LINKED ISSUE #68: Different way of checking if browser is defined?
I found a peculiar bug with using webextension-polyfill in webpack: [issue 5828](https://github.com/webpack/webpack/issues/5828).

Long story short, the polyfill guard confuses webpack in certain circumstances:
https://github.com/mozilla/webextension-polyfill/blob/8914541b8135b2e154762b01c26ee16f5a064954/src/browser-polyfill.js#L9

This got me thinking, any reason why something simpler like `if (!window.browser) {` is not used here instead?

HaNdTriX:
By checking window.browser the `webpack.ProvidePlugin` won’t be confused.

### DIFF
--- a/src/browser-polyfill.js
+++ b/src/browser-polyfill.js
@@ -6,7 +6,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 "use strict";
 
-if (typeof browser === "undefined") {
+if (typeof window.browser === "undefined") {
   // Wrapping the bulk of this polyfill in a one-time-use function is a minor
   // optimization for Firefox. Since Spidermonkey does not fully parse the
   // contents of a function until the first time it's called, and since it will


HaNdTriX:
Any progress on this?

HaNdTriX:
@rpl is there anything I can help you with to get the PR merged?

HaNdTriX:
Just in case anyone is interested in a way to temporarily solve this issue without forking this package.
You can use the [string-replace-loader](https://github.com/Va1/string-replace-loader) module to monkey patch webextension-polyfill on the fly.

```js
{
  test: /webextension-polyfill\/dist\/browser-polyfill\.js$/,
  loader: 'string-replace-loader',
  query: {
    search: 'typeof browser === "undefined"',
    replace: 'typeof window.browser === "undefined"'
  }
}
```

src/browser-polyfill.js
```diff
@@ -6,7 +6,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 "use strict";
 
-if (typeof browser === "undefined") {
+if (typeof window.browser === "undefined") {
```
ExE-Boss: Quoting myself from https://github.com/mozilla/webextension-polyfill/issues/68#issuecomment-388543539:
> I personally prefer using:
> ```js
> if (!("browser" in window)) {
>     // [...]
> }
> ```
> because using `window.browser` directly when testing if it exists tends to print dereferencing an `undefined` reference warnings in the console.

Rob--W:
The polyfill is supposed to be a no-op (not doing anything) in Firefox. In a content script, `window.browser` is `undefined` in Firefox, so this PR would incorrectly cause the global browser object to be replaced in Firefox.

Would https://github.com/mozilla/webextension-polyfill/pull/86#issuecomment-376957953, using `this.browser` instead of `window.browser` work? So that the output becomes something like this:

```javascript
if (this.browser === undefined && window.browser === undefined) {
}
```

I'm closing the issue anyway, since there is a merge conflict and we intend to change the detection logic for Edge support (#3).

vitalets:
So what is the suggested solution for using with `webpack.ProvidePlugin`?
The following does not work for me:
```js
 plugins: [
    new webpack.ProvidePlugin({
      browser: ['webextension-polyfill'],
    }),
 ]
```

HaNdTriX:
We are using the polyfill with webpack in the following way:
https://github.com/webextension-toolbox/webextension-toolbox/blob/master/src/webpack-config.js#L116-L134

ExE-Boss:
@HaNdTriX
> We are using the polyfill with webpack in the following way: https://github.com/webextension-toolbox/webextension-toolbox/blob/master/src/webpack-config.js#L116-L134
> ```js
> // The webextension-polyill doesn't work well with webpacks ProvidePlugin.
> // So we need to monkey patch it on the fly
> // More info: https://github.com/mozilla/webextension-polyfill/pull/86
> config.module.rules.push({
>   test: /webextension-polyfill[\\/]+dist[\\/]+browser-polyfill\.js$/,
>   loader: require.resolve('string-replace-loader'),
>   query: {
>     search: 'typeof browser === "undefined"',
>     replace: 'typeof window.browser === "undefined"'
>   }
> });
> ```

That won’t work on Firefox in content scripts, as in content scripts on Firefox, `this !== window`, it’s why #114 took me so long to get right.

vitalets:
@HaNdTriX Thanks for solution.

@Rob--W What about adding a short note about incompatibility with `webpack.ProvidePlugin` to the readme? It's not obvious and may take a long time to discover why it's not working. As from [webpack docs](https://webpack.js.org/plugins/provide-plugin/) Provide plugin is equivalent to `require` / `import`.

Rob--W:
@vitalets The solution from @HaNdTriX should not be used, because 1) it causes the polyfill to load unconditionally in Firefox, and 2) it causes the polyfill to break in web pages that trigger registration of the "browser" global (see #153).

I don't have a solution ready yet, but for the time being I have opened #156 to track documentation of this feature.

HaNdTriX:
@Rob--W Thanks for letting me know. Nevertheless

1. We check inside webpack and [exclude](https://github.com/webextension-toolbox/webextension-toolbox/blob/master/src/webpack-config.js#L116) the polyfill for firefox anyway.
2. Good catch! I'll try to find a workaround.

NN---:
@HaNdTriX 

```
@rpl is there anything I can help you with to get the PR merged?
What are the cons of using window.browser?
```

In Firefox window.browser is undefined as opposed to browser.
https://bugzilla.mozilla.org/show_bug.cgi?id=1208775

Issue #87, opened 2018-01-23
README.md example is misleading


eggonzal:
The example says `browser.storage.get` but in firefox is `browser.storage.local.get` is this stating that polyfill creates a different API than the actual browser API?

Rob--W:
This is a typo. The corrected version is at #90 .

rpl:
Closing. Fixed in #90.

PR #90, opened 2018-01-29
fix: Use browser.storage.local in README (#87)

Rob--W:


JimMorrison723:
you should remove the extra `);` too.

Rob--W:
@JimMorrison723 Thanks for bringing it up. I have fixed these too and verified that all code snippets are syntactically valid.

PR #94, opened 2018-02-24
Update api-metadata.json  tabs.discard

z543:
add tabs.discard


Rob--W:
Thanks for your contribution. The `discard` method was added (at the correct alphabetical location) as a part of #125, so this PR can be closed.

PR #112, opened 2018-04-17
An attempt at handling errors in message passing (perhaps not the best approach)

james-cnz:
This is my attempt at handling errors between onMessage and sendMessage.
It's perhaps not the best approach, because it duplicates a lot of code, so you might not want to use it as is, but perhaps there's something useful in it to inform a better approach?

### DIFF
--- a/src/browser-polyfill.js
+++ b/src/browser-polyfill.js
@@ -101,8 +101,8 @@ if (typeof browser === "undefined") {
       return (...callbackArgs) => {
         const lastError = chrome.runtime.lastError;
         if (lastError) {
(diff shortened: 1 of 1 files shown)


james-cnz:
There's discussion about this issue in #97  (and this patch might conflict with that patch?)

Also, I wonder if the test that failed after the spacing and comma were fixed might not have really been checking for desired behaviour?

  1) browser-polyfill wrapped runtime.onMessage listener sends the returned value as a message response:
     AssertionError: sendResponse callback has been called with the expected parameters: expected { Object (name, message) } to equal 'fake error 3'
      at thirdResponse.catch.err (test/test-runtime-onMessage.js:193:11)
      at process._tickCallback (internal/process/next_tick.js:109:7)

If I understand this right, if onMessage catches a thrown string, it is expected to call sendResponse with that string?  If so, I don't think this is really desired behaviour?

PR #114, opened 2018-05-13
Add support for browsers which don’t support Promises

### LINKED ISSUE #3: Compatibility with environments that have the `browser` namespace but don't support promises
This polyfill seems to be incompatible with Microsoft Edge, which doesn't support promises yet, but is using the `browser` instead of the `chrome` namespace.

### LINKED ISSUE #68: Different way of checking if browser is defined?

ExE-Boss:
Fixes #3
Closes #43 and closes #61

<strike>Also ~~fixes~~ #68 and ~~closes~~ #86</strike>

---

Depends on ~~#139~~ (merged)

review?(@Rob--W, @rpl)


ExE-Boss:
The build seems to be failing for reasons unrelated to the patch.

ExE-Boss:
needinfo?(@rpl)

rpl:
@ExE-Boss it is unfortunate that the integration tests are not currently able to show the actual reason for the failure in this case, but the failure is definitely related to the changes applied by this PR.

Looking if `browser` is defined on the `window` object is not currently a viable option, because it breaks the detection of the existent browser API object on Firefox, on which `window !== this` in a content script (as I also mentioned in https://github.com/mozilla/webextension-polyfill/pull/115#issue-187922657), unfortunately the test suite is not currently able to catch the actual error which is: 

```
TypeError: target is not an object or null  browser-polyfill.js:907:27
```

ExE-Boss:
It works now on Travis 🎉

ExE-Boss:
I got the build to pass on Travis 🎉

Thanks @rpl for your help 👍

ExE-Boss:
I got the build to pass on Travis 🎉

Thanks @rpl for your help 👍🏻

ExE-Boss:
I got the build to pass on Travis 🎉

src/browser-polyfill.js
```diff
@@ -6,7 +6,58 @@
...
+      try {
+        // Microsoft Edge doesn’t support `browser.runtime.getPlatformInfo()`
+        supportsPromises |= isThenable(window.browser.windows.get(browser.windows.WINDOW_ID_CURRENT));
+      } catch (e2) {
+        // Do nothing.
+      }
+    }
+  }
+  return supportsPromises;
+})()) {
```
bfred-it: That's a bit wordy, how about this function instead?

```js
function supportsPromises() {
  if (typeof browser === "undefined") {
    return false;
  }
  // If `browser.runtime.lastError` doesn’t exist, assume promises are supported.
  if (browser.runtime && typeof browser.runtime.lastError !== "undefined") {
    return true;
  }
  if (browser.extension && typeof browser.extension.lastError !== "undefined") {
    return true;
  }
  try {
    return isThenable(browser.runtime.getPlatformInfo());
  } catch (error) {}
  try {
    return isThenable(browser.windows.get(browser.windows.WINDOW_ID_CURRENT));
  } catch (error) {}
  return false;
}

if (supportsPromises()) {
```

ExE-Boss: `browser.runtime && typeof browser.runtime.lastError !== "undefined"` returns `true` if `browser.runtime.lastError` exists and `false` if it doesn’t.

bfred-it: That’s a 1-char change away, the structure remains

bfred-it: Just change `!` to `=`

bfred-it: > consider the check failed if browser.runtime.lastError exists, but browser.extension.lastError doesn’t for whatever reason.

That’s not what the current code does. The ternary operator checks for one `runtime` and then discards `extension`

ExE-Boss: Well, yours discards the one in `extension` if none exists in `runtime`.

ExE-Boss: <i title="More sarcasm">Oh great</i>, and now Firefox [stopped working](https://travis-ci.org/mozilla/webextension-polyfill/builds/393441006), again.

ExE-Boss:
@bfred-it <i title="More sarcasm">Oh great</i>, and now Firefox [stopped working](https://travis-ci.org/mozilla/webextension-polyfill/builds/393441006), again.

src/browser-polyfill.js
```diff
@@ -6,7 +6,48 @@
...
+    return true;
+  }
+  if (browser.extension && typeof browser.extension.lastError === "undefined") {
+    return true;
+  }
+  try {
+    return isThenable(browser.runtime.getPlatformInfo());
+  } catch (error) {
+    // Microsoft Edge doesn’t support `browser.runtime.getPlatformInfo()`
+    try {
```
bfred-it: There’s not need to nest this. If there are no errors, the function will return before the second try/catch

ExE-Boss:
I did it, it works correctly now in all browsers and doesn’t wrap the existing browser object in Firefox.

ExE-Boss:
> With the above changes, we are making sure that `extensionAPIs` is going to be the original "extensions APIs" object and so we will also need to change the two places where the polyfill is currently referring directly to `chrome.runtime.lastError`, so that it uses `extensionAPIs.runtime.lastError`

It already does that by making `chrome` a parameter to the `wrapAPIs` function: https://github.com/mozilla/webextension-polyfill/blob/a62cf867304081c6245664355842c7663f5da2e2/src/browser-polyfill.js#L60

rpl:
> It already does that by making chrome a parameter to the wrapAPIs function:

yeah I did notice that, but I'm thinking (aloud) that I would kind of prefer if we don't use `chrome` elsewhere inside the polyfill's "wrapping code" (besides the initial preamble), so that it cannot be mistaken as accessing the global `chrome` object while reading the code 
(I mean, clearly it doesn't make any difference for the machines, but it may be helpful for humans ;-)).

ExE-Boss:
> > It already does that by making `chrome` a parameter to the `wrapAPIs` function:
>
> yeah I did notice that, but I'm thinking (aloud) that I would kind of prefer if we don't use `chrome` elsewhere inside the polyfill's "wrapping code" (besides the initial preamble), so that it cannot be mistaken as accessing the global `chrome` object while reading the code.

The reason why I didn’t rename `chrome` to `extensionAPIs` was so that other PRs wouldn’t need to be rebased and to minimise possible merge conflicts (I’ve opened PR #139 to make this change).

NN---:
Any updates?
This repository seems not so maintained what makes people to invent a different workarounds.

src/browser-polyfill.js
```diff
@@ -6,7 +6,52 @@
...
+ * Checks if the current browser supports
+ * the Promise-based WebExtension APIs.
+ *
+ * @returns {boolean} True if yes.
+ */
+const supportsPromises = () => {
+  if (!HAS_BROWSER_NAMESPACE) {
+    return false;
+  }
+  // If `browser.runtime.lastError` doesn’t exist, assume promises are supported.
```
leonid-shevtsov: In Edge, these two checks don't work. The `browser.runtime.lastError` property is present, but it's `undefined` in the absence of an error. Maybe a check for `hasOwnProperty("lastError")` would be more appropriate, but I am not an expert.

However, `browser.extension.lastError` is not defined at all, even though Edge does not support promises and needs the polyfill.

leonid-shevtsov:
Also, `window.browser = require('browser-polyfill')` messes up the proxying wrappers and produces weird errors. Which is fine - you just need to stop using the global `browser` variable.

StigNygaard:
Is it okay to ask if there's still development on this, and if we can expect "unofficial support" for MS Edge in a foreseeable future?

Gaelan:
Edge is becoming a Chromium variant, so this might be pointless.

ExE-Boss:
@Gaelan
> Edge is becoming a Chromium variant, so this might be pointless.

This PR isn’t just for **Microsoft&nbsp;Edge** (even though that is the main reason why this PR got made).

This PR is to support browsers which define the `browser` namespace but don’t support using promises, and instead require callbacks.

Also, even if Edge will switch to Chromium, there’s still the fact that if Microsoft wants to be backwards compatible, they’ll have to keep defining the `browser` namespace.

In fact, Edge doesn’t support the `chrome` namespace at all, unlike Firefox, which supports it for compatibility purposes.

Rob--W:
Since the announcement of Microsoft that Edge switches to Chromium, the effort to support Edge is put on hold until the exact details become available. It is quite likely for Edge to behave identical to Chrome, in which case we don't need any more modifications to support Edge.

> This PR is to support browsers which define the `browser` namespace but don’t support using promises, and instead require callbacks.

I'm not aware of such a browser. Let's keep the main library simple; if anyone wants to develop an extension for non-Chromium Edge browsers, then they could create a build based on this pull request.

NN---:
@Rob--W The latest Windows 1809 is [supported ](https://support.microsoft.com/en-us/help/13853/windows-lifecycle-fact-sheet) till May 11, 2021.
The library is needed for anyone using this version.
As well it is not clear whether WIndows 1903 will be released with Edge based Chromium moving support lifecycle even further.

Current solution/workaround is to use 'chrome' object and not 'browser'.
Write code with callbacks and not promises.
And for Edge just add one line:   window.chrome = window.browser.
That way one can have one code base for Chrome,FF,Edge.

But this is not what people want. We want Promises API instead of callbacks which work in all browsers.

Rob--W:
> @Rob--W The latest Windows 1809 is supported till May 11, 2021.
> The library is needed for anyone using this version.
> As well it is not clear whether WIndows 1903 will be released with Edge based Chromium moving support lifecycle even further.

"Microsoft Edge will now be delivered and updated for all supported versions of Windows" ([source](https://blogs.windows.com/windowsexperience/2018/12/06/microsoft-edge-making-the-web-better-through-more-open-source-collaboration/)). This implies that supported versions of Windows (including 1809) will receive a Chromium-based Edge browser.

> Except that Chromium Edge will certainly include the `browser` namespace due to Microsoft’s obsession with backwards compatibility.

I'd like to see a reputable source that supports this claim.

> Well, Edge matches that description, and Chromium Edge most certainly will too since normal Chromium doesn’t yet use Promises for the extension API.

By "I'm not aware of such a browser", I meant that there is no supported browser (Microsoft is discontinuing their current engine, so that doesn't count) that would get API support through this PR. The polyfill already supports Chromium-based browsers.

rpl:
As @Rob--W already mentioned in his last round of comments, with the Microsoft announcement related to the MSEdge future there are even less compelling reasons to officially include in the polyfill any workarounds specific to MSEdge (and then maintaining them over the time).

There are no other browsers that exposes the `browser` global, and personally I was already not thrilled at all about adding to the polyfill "runtime detection of the Promised-based APIs".

MSEdge doesn't have a lot of extensions, and it is very likely that the extensions that they have are already been developed for Chrome, and so I'm not really sure that they need to define the `browser` global to keep any compatibility for them, in my personal opinion their best "compatibility"-move is more likely to just "try to don't be incompatible with existing Chrome and Firefox extensions".

For anyone that still have to work on MSEdge extensions before it is switched to the next Chromium-based version, a better bet may be to do the tweaks needed to make "MSEdge to look like Chromium (and allow the webextension-polyfill to load unmodified)" using the MSEdge specific `--ms-preload` manifest key and the [Microsoft Edge Extension Toolkit](https://docs.microsoft.com/en-us/microsoft-edge/extensions/guides/porting-chrome-extensions)'s Chrome API bridges, I've just created the following github repository to explain this alternative approach in a bit more detail and provide a minimal test extension that shows how to put it together:

- https://github.com/rpl/example-msedge-extension-with-webextension-polyfill

I'm going to close this PR, #163 and #3 as wontfix, for the reasons described above and in the last round of @Rob--W comments (and I'm not say this lightly, from my point of view it is quite sad that we are losing an additional indipendent implementation of the Extension APIs, and I was really willing to add MSEdge support as much as I could, e.g. that was the reason why I've been working on applying all the changes needed to run all our integration tests on MSEdge in #163), and I'm going to update the 'Supported Browsers' section from the README.md file in this repo accordingly.

PR #115, opened 2018-05-14
test: Add cross browsers smoke tests and add support for sendMessage rejected promise behavior   

rpl:
This pull request applies some changes to the "smoke tests", which currently runs only on Chrome using the puppeteer npm package, to make them run on more then one browser (currently Chrome and Firefox).

The new approach is collecting test results from the test extensions, in the "tap protocol" format, and merge them into a single "per browser" test suite, then run the entire test suite on different browsers.

The supported browsers (currently Chrome and Firefox) are launched and controller using selenium-webdriver (which [at least theoretically should be allow us to support also Edge](https://docs.microsoft.com/en-us/microsoft-edge/extensions/guides/packaging/creating-and-testing-extension-packages#automated-testing-with-webdriver), but I've been currently unable to make it works on Edge as documented in the Microsoft doc page).

On top of the changes needed on the test suite, there are the following changes:

- Print a deprecation warning (only once) when the deprecated sendResponse callback has been called
  (we added it because the current behavior, defined and working on Firefox and undefined on Chrome, was confusing, but it is still true that it should be deprecated and removed once it has been removed from the W3C draft and from Firefox)

-  Reject the promised returned by sendMessage, when a onMessage listener returns a rejected promise (as close as possible to the behavior implemented natively on Firefox)

- Test explicitly that the polyfill "existing browser API object" detection is not wrapping the chrome API namespace on Firefox by mistake (which shows practically why the simple change proposed in #68 / #86 / #114 , in a Firefox content script `this !== window` and the proposed check is going to make the polyfill to wrap the chrome API object and overwrite the native browser API object with the polyfill one)


src/browser-polyfill.js
```diff
@@ -382,36 +396,96 @@ if (typeof browser === "undefined") {
...
         // Let Chrome know that the listener is replying.
         return true;
       };
     });
 
+    const wrappedSendMessageCallback = ({reject, resolve, unexpectedError}, reply) => {
+      if (reply && reply.__webExtensionPolyfillReject__) {
+        const {error} = reply;
+        // Convert the JSON representation of the error back to an Error instance,
+        // otherwise just reject with the error value as it is.
```
james-cnz: This comment line "// otherwise just reject with the error value as it is." no longer applies?

src/browser-polyfill.js
```diff
@@ -382,36 +396,96 @@ if (typeof browser === "undefined") {
...
+            console.error(error);
+            // Send a JSON representation of the error if the rejected value
+            // is an instance of error, or the object itself otherwise.
+            const isErrorInstance = error instanceof Error;
+            sendResponse({
+              __webExtensionPolyfillReject__: {isErrorInstance},
+              error: !isErrorInstance ? error : {
+                message: error.message,
+                fileName: error.fileName,
+                lineNumber: error.lineNumber,
```
james-cnz: I think fileName and lineNumber are non-standard, and only supported by Firefox, which doesn't need the polyfill anyway, so there's no point trying to use them?

src/browser-polyfill.js
```diff
@@ -382,36 +396,96 @@ if (typeof browser === "undefined") {
...
       };
     });
 
+    const wrappedSendMessageCallback = ({reject, resolve, unexpectedError}, reply) => {
+      if (reply && reply.__webExtensionPolyfillReject__) {
+        const {error} = reply;
+        // Convert the JSON representation of the error back to an Error instance,
+        // otherwise just reject with the error value as it is.
+        if (reply.__webExtensionPolyfillReject__.isErrorInstance) {
+          reject(new Error(error.message, error.fileName, error.lineNumber));
```
james-cnz: As before, I don't think fileName or lineNumber are usable? Would it instead make sense to always throw unexpectedError, after setting its message property if a message is available?

src/browser-polyfill.js
```diff
@@ -382,36 +396,96 @@ if (typeof browser === "undefined") {
...
+        // to translate the message into a resolved promise or a rejected
+        // promise).
+        const sendPromisedResult = (promise) => {
+          promise.then(msg => {
+            // send the message value.
+            sendResponse(msg);
+          }, error => {
+            console.error(error);
+            // Send a JSON representation of the error if the rejected value
+            // is an instance of error, or the object itself otherwise.
```
james-cnz: This last bit of the comment "or the object itself otherwise" no longer applies?

src/browser-polyfill.js
```diff
@@ -382,36 +396,96 @@ if (typeof browser === "undefined") {
...
+        if (reply.__webExtensionPolyfillReject__.isErrorInstance) {
+          reject(new Error(error.message, error.fileName, error.lineNumber));
+        } else {
+          // If the original error wasn't an error instance, let's try to keep
+          // the behavior as close as possible to the Firefox behavior, e.g.
+          // print the error object as received in the console and then reject
+          // using a generic error object.
+          console.error(error);
+          reject(unexpectedError);
+        }
```
james-cnz: I suggest replacing the above 13 lines with:
// Convert the JSON representation of the error back to an Error instance.
`unexpectedError.message = reply.message;`
`console.error(unexpectedError);`
`reject(unexpectedError);`

src/browser-polyfill.js
```diff
@@ -396,36 +396,96 @@ if (typeof browser === "undefined") {
...
+      }
+
+      if (args.length > metadata.maxArgs) {
+        throw new Error(`Expected at most ${metadata.maxArgs} ${pluralizeArguments(metadata.maxArgs)} for ${name}(), got ${args.length}`);
+      }
+
+      // Create an error object here, so that the original caller is part of the
+      // stack trace (similarly to the stack of the error raised on Firefox in
+      // this scenario).
+      const unexpectedError = new Error("An unexpected error occurred");
```
Rob--W: I'm missing a test case that checks this behavior.

src/browser-polyfill.js
```diff
@@ -396,36 +396,96 @@ if (typeof browser === "undefined") {
...
+        const {error} = reply;
+        // Convert the JSON representation of the error back to an Error instance,
+        // otherwise just reject with the error value as it is.
+        if (reply.__webExtensionPolyfillReject__.isErrorInstance) {
+          reject(new Error(error.message, error.fileName, error.lineNumber));
+        } else {
+          // If the original error wasn't an error instance, let's try to keep
+          // the behavior as close as possible to the Firefox behavior, e.g.
+          // print the error object as received in the console and then reject
+          // using a generic error object.
```
Rob--W: This is not Firefox's behavior, and even if it was, then it would be too implementation-specific.

Test case (Firefox Nightly 61, buildID 20180504220115):
```javascript
browser.runtime.onMessage.addListener(()=>Promise.reject({message:"what"}))
```

```javascript
// If doing this, no error printed:
browser.runtime.sendMessage('ee').catch(()=>{});
// If doing this, then error is printed (because it's an uncaught promise rejection):
browser.runtime.sendMessage('ee');
```

src/browser-polyfill.js
```diff
@@ -396,36 +396,96 @@ if (typeof browser === "undefined") {
...
         return true;
       };
     });
 
+    const wrappedSendMessageCallback = ({reject, resolve, unexpectedError}, reply) => {
+      if (reply && reply.__webExtensionPolyfillReject__) {
+        const {error} = reply;
+        // Convert the JSON representation of the error back to an Error instance,
+        // otherwise just reject with the error value as it is.
+        if (reply.__webExtensionPolyfillReject__.isErrorInstance) {
```
Rob--W: Let's only forward the message. One cannot rely on `fileName` / `lineNumber` (and it is not even documented).

test/fixtures/runtime-messaging-extension/background.js
```diff
@@ -33,6 +33,12 @@ browser.runtime.onMessage.addListener((msg, sender, sendResponse) => {
       // a reply from the second listener.
       return false;
 
+    case "test - sendMessage with returned rejected Promise with Error value":
+      return Promise.reject(new Error("rejected-error-value"));
+
+    case "test - sendMessage with returned rejected Promise with non-Error value":
+      return Promise.reject("rejected-non-error-value");
```
Rob--W: Try `Promise.reject({message: "rejected-non-error-value"})` instead.

If you wish, keep the existing test case, but check that no useful error message is passed through.

test/fixtures/tape-standalone.js
```diff
@@ -0,0 +1,39 @@
...
+
+let browser = "unknown";
+if (navigator.userAgent.includes("Chrome/")) {
+  browser = "Chrome";
+} else if (navigator.userAgent.includes("Firefox/")) {
+  browser = "Firefox";
+}
+
+// Export as a global a wrapped test function which enforces a timeout by default.
+window.test = (desc, fn) => {
```
Rob--W: To make it clear that `test` is not a regular tap `test`, use a different name, e.g. `describeBrowserTest`.

src/browser-polyfill.js
```diff
@@ -382,36 +396,96 @@ if (typeof browser === "undefined") {
...
+            // is an instance of error, or the object itself otherwise.
+            const isErrorInstance = error instanceof Error;
+            sendResponse({
+              __webExtensionPolyfillReject__: {isErrorInstance},
+              error: !isErrorInstance ? error : {
+                message: error.message,
+                fileName: error.fileName,
+                lineNumber: error.lineNumber,
+              },
+            });
```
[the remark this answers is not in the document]
ExE-Boss: I personally prefer `__mozWebExtPolyfill_isSerializedErrorObject` over `__webExtensionPolyfillReject__`, [the reasons are explained here](https://github.com/mozilla/webextension-polyfill/pull/97#discussion_r178991820):
> Perhaps something like `__mozWebExtPolyfill_isSerializedErrorObject` could work, because `isInstanceOfError` is likely to be used by something.
>
> The double-`_` prefix is used since it’s somewhat common practice to use `__comment` in place of real comments in JSON, and a single-`_` prefix already denotes an internal varible.
>
> The [`moz` prefix](https://developer.mozilla.org/en-US/docs/Glossary/Vendor_Prefix#Property_and_method_prefixes) is also used by non-standard Mozilla properties on Web APIs, and `WebExtPolyfill` identifies this polyfill.

test/integration/setup.js
```diff
@@ -1,9 +1,69 @@
...
+  const browser = launchOptions.browser || process.env.TEST_BROWSER_TYPE;
+  const extensionPath = launchOptions.extensionPath;
+
+  let driver;
+
+  if (browser === "chrome") {
+    const options = new chrome.Options();
+    options.addArguments([
+      `--load-extension=${extensionPath}`,
+      "--no-sandbox",
```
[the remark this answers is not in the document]
ExE-Boss: The comment should say: <code>// See issue <a href="https://github.com/mozilla/webextension-polyfill/pull/85" title="#85, fix: Pass --no-sandbox chrome CLI option when running the integration tests on travis">#85</a></code>

test/fixtures/runtime-messaging-extension/background.js
james-cnz: If keeping the existing test case, maybe check specifically for an Error with the message "An unexpected error occurred"?

src/browser-polyfill.js
james-cnz: I've made a suggestion for this, see: <a href="https://github.com/mozilla/webextension-polyfill/pull/115#pullrequestreview-120089290">onMessage side</a> and <a href="https://github.com/mozilla/webextension-polyfill/pull/115#pullrequestreview-120089863">sendMessage side</a>.

src/browser-polyfill.js
james-cnz: If I understand <a href="https://github.com/mozilla/webextension-polyfill/pull/115/files#r188247571">this comment re console error messages</a> correctly, maybe remove the console error line?

test/fixtures/runtime-messaging-extension/background.js
james-cnz: For what it's worth, I think this mistake was already present in the code, not introduced by the pull request.

.travis.yml
```diff
@@ -2,7 +2,7 @@ language: node_js
 sudo: false
 node_js:
 ## Some of the ES6 syntax used in the browser-polyfill sources is only supported on nodejs >= 6
-- '6'
+- '8'
```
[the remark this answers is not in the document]
rpl: nodejs 8 is needed to load successfully some of the selenium-webdriver modules (I could theoretically force nodejs 6 to load them by configuring babel-register to transpile the modules loaded from node_modules, but I'm not sure that it is worth it)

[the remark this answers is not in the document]
rpl: (In particular this one is the first to fail to load: 
```
/.../webextension-polyfill/node_modules/selenium-webdriver/lib/http.js:454
  async execute(command) {
        ^^^^^^^

SyntaxError: Unexpected identifier
    at createScript (vm.js:56:10)
    at Object.runInThisContext (vm.js:97:10)
    at Module._compile (module.js:549:28)
    at Module._extensions..js (module.js:586:10)
    at Object.require.extensions.(anonymous function) [as .js] (/.../webextension-polyfill/node_modules/babel-register/lib/node.js:152:7)
    at Module.load (module.js:494:32)
    at tryModuleLoad (module.js:453:12)
    at Function.Module._load (module.js:445:3)
    at Module.require (module.js:504:17)
    at require (internal/module.js:20:19)
```
)

.travis.yml
```diff
@@ -14,7 +14,13 @@ script:
 - export DISPLAY=:99.0
 - sh -e /etc/init.d/xvfb start
 - echo "RUN integration tests on chrome" &&
-  TRAVIS_CI=true ./test/run-chrome-smoketests.sh
+  TRAVIS_CI=true ./test/run-browsers-smoketests.sh
+
+sudo: required
```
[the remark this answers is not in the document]
rpl: To be fair I've not yet tried to remove it, but travis doc page related to chrome was suggesting it: https://docs.travis-ci.com/user/chrome

src/browser-polyfill.js
```diff
@@ -382,36 +396,96 @@ if (typeof browser === "undefined") {
...
-          });
+          sendPromisedResult(sendResponsePromise);
         }
 
         // Let Chrome know that the listener is replying.
         return true;
       };
     });
 
+    const wrappedSendMessageCallback = ({reject, resolve, unexpectedError}, reply) => {
```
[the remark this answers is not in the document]
rpl: That's a good point, I'm not sure that is going to be effective (and I'm not sure how we could trigger it in one of the test cases), but at least it should not do any harm (chrome should already guarantees that if there is an error set on chrome.runtime.lastError it should be related to the currently executed callback).

@Rob--W do you have any opinion (or specific knowledge) about this?

src/browser-polyfill.js
rpl: :+1: done (as briefly described in one of the above comments)

src/browser-polyfill.js
rpl: :+1: I was also pretty unsure about these console errors, to be fair I'm still a bit worried that the stack traces that the extension is going to be able to access directly from their code could be not that helpful (at least in some of the scenarios), but I've removed all the redundant console messages for now.

src/browser-polyfill.js
rpl: for now I've removed that part (and the console.error).

src/browser-polyfill.js
rpl: This particular line is gone in the updated version.

A test case for the "An unexpected error occurred" is in the current set of tests but like I briefly explain in the inline comment, Firefox seems to present a different behavior in version 59 (where the error has the "An unexpected error occurred" message) and 60 (where it seems that it is currently "undefined"), and so the test cases is not currently asserting the error message (it just checks that an instance of Error is rejected in that case).

test/fixtures/runtime-messaging-extension/background.js
rpl: I added both the cases, to be fair I think that the behavior that Firefox have when an extension code rejects a plain object with a message property is actually an involuntary side-effect of the behavior we wanted when an API implementation rejects that kind of plain objects: 

https://searchfox.org/mozilla-central/search?q=Promise.reject(%7Bmessage%3A&case=false&regexp=false&path= 

Anyway, the polyfill is currently matching this behavior and I added an explicit test case for it.

@james-cnz Like I described in one of the above comments I'm not asserting the "An unexpected error occurred" error message because Firefox 59 and Firefox 60 have different behaviors (and so the test was failing on travis only on Firefox).

test/fixtures/tape-standalone.js
rpl: uhm... not sure that it is worth the additional verbosity in the test files, in the end it is just a wrapper around tape which ensure that the test will timeout (because it wasn't and I would really prefer that it does timeout at some point if no message is being received, so that we can more easily guess where the test was failing directly from the failures on travis).

test/integration/setup.js
```diff
@@ -1,9 +1,69 @@
...
 const serveStatic = require("serve-static");
-const puppeteer = require("puppeteer");
 
-exports.createHTTPServer = async (path) => {
+const geckodriver = require("geckodriver");
+const chromedriver = require("chromedriver");
+const {Builder, By, until} = require("selenium-webdriver");
+const {Command} = require("selenium-webdriver/lib/command");
+const chrome = require("selenium-webdriver/chrome");
+const firefox = require("selenium-webdriver/firefox");
```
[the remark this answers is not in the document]
rpl: :+1: I moved the require for the browser-specific ones in the related launchBrowser if-blocks.

test/integration/setup.js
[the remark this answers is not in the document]
rpl: :+1 done

test/integration/setup.js
```diff
@@ -1,9 +1,69 @@
...
+    options.addArguments([
+      `--load-extension=${extensionPath}`,
+      "--no-sandbox",
+    ]);
+
+    driver = await new Builder()
+      .forBrowser("chrome")
+      .setChromeOptions(options)
+      .setChromeService(new chrome.ServiceBuilder(chromedriver.path))
+      .build();
```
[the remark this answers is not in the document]
rpl: uhm, yeah, sounds good to me

rpl:
@ExE-Boss @james-cnz @Rob--W Thank you so much for all your review comments, you have been really really helpful!

This should not be ready for another round of review.

src/browser-polyfill.js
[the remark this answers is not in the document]
Rob--W: An error could be triggered if you send a message without recipient. E.g. `chrome.tabs.sendMessage(SOME_NON_EXISTENT_TAB_ID, msg, callback);`

Note that unlike many other extension APIs, Chrome does not print an unchecked error to the console (but `chrome.runtime.lastError` is set). (If you try to test this in Chrome and can't confirm this behavior, then you might be affected by this bug: https://bugs.chromium.org/p/chromium/issues/detail?id=836370 )

test/integration/setup.js
```diff
@@ -101,8 +101,8 @@ async function runExtensionTest(t, server, driver, extensionDirName) {
 
     // Merge tap results from the connected browser.
     const el = await driver.wait(until.elementLocated(By.id("test-results")), 10000);
-    const tests = (await el.getAttribute("textContent")).split("\n");
-    console.log(tests.filter(row => !row.startsWith("TAP version")).join("\n"));
+    const testResults = (await el.getAttribute("textContent"));
```
Rob--W: Unnecessary braces around `await ...`.

test/fixtures/tape-standalone.js
Rob--W: If it was a real `test` from tap, then `test.skip` would also work. But we don't use it yet, so if you prefer, we can keep `test`.

ExE-Boss:
Something in this broke the #114 build, but in a way which isn’t caused by #114.

PR #122, opened 2018-06-01
feat: Tool to automatically collect API information from Chrome

Rob--W:
This tool automatically generates metadata from Chrome's runtime, by walking the API namespace, calling the methods with invalid parameters and catching the error. The error message includes a useful method signature.

This PR contains two test extensions, because the `page_action` and `browser_action` manifest keys are mutually exclusive.  I have added a Node.js server to collect and merge the results from separate runs.

The permissions to unlock the APIs are based on:

- https://developer.chrome.com/extensions/api_index
- https://developer.chrome.com/extensions/permission_warnings#permissions_with_warnings

I double-checked with:

- https://chromium.googlesource.com/chromium/src/+/2f2f1153ea9f8e6c5f277ee5a7cbfc242bf6d3ce/extensions/common/api/_permission_features.json

Alternatives that I have considered and rejected:

- Parsing the schemas directly from the source code. Chromium's API definitions are scattered over JSON and IDL files (for schemas) and also some other files for permissions and availability (JSON). I didn't want to write a new parser for it. The source also contains Python scripts that parse everything, but re-using to get our desired information takes some efforts (and a multi-GB Chromium checkout).
- Scraping the schemas from Chrome's documentation: Rejected because the documentation is not consistently updated: https://bugs.chromium.org/p/chromium/issues/detail?id=513780#c23

### DIFF
--- a/tools/metadata-generator/README.md
+++ b/tools/metadata-generator/README.md
@@ -1,0 +1,64 @@
+This tool collects the API namespaces from Chrome.
+
+### Usage
+
+1. Start `./collector.js`
+2. Start Chrome and load the two extensions (there are two extensions because
+   it is not possible to have one extension with all APIs).
+3. Start Chrome again, with `--enable-features=NativeCrxBindings` and load one
+   of the two extensions.
+4. To support older versions of Chrome (which may have different API expectations),
+   repeat step 3 for each version of Chrome.
+
+When you are done, the collected metadata is available at http://127.0.0.1:55781/
+
+To only see the namespaces that already exist in api-manifest.json, visit
+http://127.0.0.1:55781/subset
+
+Then use jsondiff, e.g. jsondiff.com to compare the old and new metadata.
+
+
+### Example
+For example, to collect the metadata from multiple sources:
+
+```sh
+node ./collector.js
+
+# Now in a different terminal:
+chromium --no-first-run --user-data-dir=/tmp/whatever --load-extension=extension-apis,extension-apis/browseraction data:,extension-apis data:,browseraction
+chromium --no-first-run --user-data-dir=/tmp/whatever --load-extension=extension-apis --enable-features=NativeCrxBindings data:,extension-apis
+
+# If repeating the last line for old Chrome versions, remove the profile dir
+# first using:   rm -r /tmp/whatever
+
+# See JSON output in stdout
+curl http://127.0.0.1:55781/subset
+
+# When you are done, go back to the terminal that runs collector.js and ^C
+```
+
+Explanation of above commands:
+
+- `--load-extension=comma-separated-list-of-directories to load extension(s).
+- `data:,extension-apis` is a data-URL.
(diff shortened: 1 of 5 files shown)


Rob--W:
For reference, the tool output is:

API subset from Chromium 54, 67, 67 with NativeCrxBindings: https://pastebin.com/raw/JdSTEN8S
The full file (including the omited namespaces) is: https://pastebin.com/raw/Sn7ctsUi

The following namespaces are excluded from the subset because they do not appear in the current namespace:

- browsingData
- debugger
- desktopCapture
- fontSettings
- gcm
- instanceID
- pageCapture
- permissions
- sessions
- system
- tabCapture
- topSites
- tts

Note that `browsingData`, `permissions` and `topSites` should have been included because Firefox supports them...

Also note that my tool does not generate anything in the `devtools` namespace because the `devtools` API is special in Chrome (implemented in JavaScript without wrappers for extension APIs).

Finally, my tool excludes the APIs in the `privacy` and `contentSettings` namespaces because these are basically copy & paste APIs, and we should do something special for them instead of duplicating exactly the same schema a zillion times for every subkey of `privacy`.

Rob--W:
I have updated the commit to include the "downloads.open" and "downloads.shelf" permission. The output is not different, because these functions don't take any callbacks.

brit878:
> This tool automatically generates metadata from Chrome's runtime, by walking the API namespace, calling the methods with invalid parameters and catching the error. The error message includes a useful method signature.
> 
> This PR contains two test extensions, because the `page_action` and `browser_action` manifest keys are mutually exclusive. I have added a Node.js server to collect and merge the results from separate runs.
> 
> The permissions to unlock the APIs are based on:
> 
> * https://developer.chrome.com/extensions/api_index
> * https://developer.chrome.com/extensions/permission_warnings#permissions_with_warnings
> 
> I double-checked with:
> 
> * https://chromium.googlesource.com/chromium/src/+/2f2f1153ea9f8e6c5f277ee5a7cbfc242bf6d3ce/extensions/common/api/_permission_features.json
> 
> Alternatives that I have considered and rejected:
> 
> * Parsing the schemas directly from the source code. Chromium's API definitions are scattered over JSON and IDL files (for schemas) and also some other files for permissions and availability (JSON). I didn't want to write a new parser for it. The source also contains Python scripts that parse everything, but re-using to get our desired information takes some efforts (and a multi-GB Chromium checkout).
> * Scraping the schemas from Chrome's documentation: Rejected because the documentation is not consistently updated: https://bugs.chromium.org/p/chromium/issues/detail?id=513780#c23

Stop

infokiller:
This looks great. Any reason it's not merged?

Rob--W:
> This looks great. Any reason it's not merged?

The tool was used once and kept here for reference. It is not actively maintained: if new APIs are added to Chrome, then the changes won't be reflected here because the permission needs to be added to the manifest file of the tool.

PR #124, opened 2018-06-04
Use yargs to allow choosing the browser using a command line parameter

ExE-Boss:
Using environment variables for testing manually is annoying.

This makes it possible to use the `--browser`/`-b` command line parameter to choose the browser.


rpl:
@ExE-Boss Thanks for this proposal, but we have discussed about it and given that it is just the entrypoint script for the integration test suite, we prefer to solve it in the docs (that now mention explicitly how to run the integration tests on a single browser) and do not add an additional dev dependency to parse the script arguments.

PR #127, opened 2018-06-17
Fix tests when installed using pnpm

ExE-Boss:
[`pnpm`](https://pnpm.js.org) is the third largest node package manager, which uses symlinks, hardlinks and isolated trees, which prevents packages from using packages which they haven’t declared in their dependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
   "homepage": "https://github.com/mozilla/webextension-polyfill",
   "devDependencies": {
     "async-wait-until": "^1.1.5",
+    "babel-core": "^6.26.3",
     "babel-eslint": "^8.0.1",
     "babel-plugin-transform-es2015-modules-umd": "^6.24.1",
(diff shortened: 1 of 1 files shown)


ExE-Boss:
**Q:** Why is this still unmerged?

Rob--W:
@ExE-Boss Tomorrow is the triage meeting where I and @rpl will go through all open PRs.

rpl:
@ExE-Boss @Rob--W actually, this PR looks already good, I didn't notice that Rob didn't merged it yet.
I also tried it locally (just because I never used pnpm before and I wanted to see when the issues were triggered, e.g. with babel-core not explicitly listed in the devDependencies makes "pnpm install" to fail during while it is building the polyfill, and the missing tape makes the integration tests to fails because it doesn't find the tape executable, unsurprisingly).

Thanks @ExE-Boss!

PR #128, opened 2018-06-17
Use `delete window.browser` in `setupTestDOMWindow`

ExE-Boss:
Now that jsdom/jsdom#1622 has been fixed, and the minimum supported Node version for running the tests is 8 (which has the relevant fix), the relevant follow‑up from #2 can be applied.


rpl:
@ExE-Boss ah! that's great! it feels great to remove a workaround when finally it is not needed anymore \o/ 

Thanks for noticing that it was fixed upstream and for taking care of creating a pull request for it :+1:

PR #129, opened 2018-06-17
Correct `.travis.yml`

ExE-Boss:
This fixes issues introduced by #115 and #66.


rpl:
@ExE-Boss it looks great, thanks for applying these cleanups!

PR #131, opened 2018-06-19
Silence irrelevant browser.runtime.sendMessage error

### LINKED ISSUE #130: Chrome: "The message port closed before a response was received."
**Minimal repro repo:**
https://github.com/lydell/webextension-polyfill-messaging-issue

This happens with both v0.2.1 on npm and when built from commit 2537b23837d4752e3a6c3236c2776adbc89f1312.

# Summary

If a `browser.runtime.onMessage` callback does not return a Promise, `browser.runtime.sendMessage` will be rejected with the following error, causing noise in the console:

> "The message port closed before a response was received."

## Workaround

Even if you don't want to send a response, always return a promise
in your `onMessage` callback:

```js
// background.js
browser.runtime.onMessage.addListener(message => {
  console.log("background: onMessage", message);

  // Add this line:
  return Promise.resolve("Dummy response to keep the console quiet");
});
```

You can also make the `onMessage` callback `async` to implicitly return a Promise (resolving to `undefined` in the below example).

```js
// background.js
// Notice the `async` keyword.
browser.runtime.onMessage.addListener(async message => {
  console.log("background: onMessage", message);
});
```

# Files

Copied over for convenience from:
https://github.com/lydell/webextension-polyfill-messaging-issue

```json
{
  "manifest_version": 2,
  "version": "0.0.0",
  "name": "Test",
  "background": {
    "scripts": [
      "browser-polyfill-npm.js",
      "background.js"
    ]
  },
  "content_scripts": [
    {
      "matches": [
        "<all_urls>"
      ],
      "js": [
        "browser-polyfill-master.js",
        "content.js"
      ]
    }
  ]
}
```

```js
// background.js

browser.runtime.onMessage.addListener(onMessage);

function onMessage(message) {
  console.log("background: onMessage", message);

  // 1: Causes the following to be logged in content:
  // "The message port closed before a response was received."
  return undefined;

  // 2: Causes this response to be logged in content, as expected.
  // return Promise.resolve("response from background");

  // 3: Causes this error to be logged in content, as expected.
  // return Promise.reject(new Error("Could not respond"));

  // 4: Causes nothing at all to be logged in content!
  // I guess it is waiting for the deprecated `sendResponse` parameter to be
  // called.
  // return true;
}
```

```js
// content.js

// 1: Unless background returns a Promise in its onMessage, this promise is
// rejected with:
// "The message port closed before a response was received."
browser.runtime
  .sendMessage("hello from content")
  .then(console.log, console.error);



// 2: This does not seem to cause any errors:
// chrome.runtime.sendMessage("hello from content");
// console.log("content: after chrome.runtime.sendMessage", chrome.runtime.lastError);



// 3: Inside the callback, `chrome.runtime.lastError` will be:
// "The message port closed before a response was received."
// It seems like if `sendMessage` defines a callback but the other end doesn't
// respond, Chrome is treating that as an error. Which makes sense.
// The question is how this should be handled in a Promise based API.
// chrome.runtime.sendMessage("hello from content", response => {
//   console.log("content: callback", response, chrome.runtime.lastError);
// });
// console.log(
//   "content: after chrome.runtime.sendMessage with callback",
//   chrome.runtime.lastError
// );
```

## Solution?

Should the "The message port closed before a response was received." be detected, and the promise should be resolved with `undefined`?

lydell:
Fixes #130.

See the comments in the code.

You can test here (by changing to `browser-polyfill-fork.js` in `manifest.json`): https://github.com/lydell/webextension-polyfill-messaging-issue

### DIFF
--- a/src/browser-polyfill.js
+++ b/src/browser-polyfill.js
@@ -13,6 +13,8 @@ if (typeof browser === "undefined") {
       removed from the specs (See
       https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/runtime/onMessage)
     `.replace(/\s+/g, " ").trim();
+  const CHROME_SEND_MESSAGE_CALLBACK_NO_RESPONSE_MESSAGE =
+    "The message port closed before a response was received.";
 
   // Wrapping the bulk of this polyfill in a one-time-use function is a minor
   // optimization for Firefox. Since Spidermonkey does not fully parse the
@@ -446,7 +448,23 @@ if (typeof browser === "undefined") {
 
     const wrappedSendMessageCallback = ({reject, resolve}, reply) => {
       if (chrome.runtime.lastError) {
-        reject(chrome.runtime.lastError);
+        // Passing a callback to `chrome.runtime.sendMessage`
+        // (`chrome.runtime.sendMessage("some_message", response => {
+        // console.log(response) })`) is a mistake if the other end does not
+        // send a response. As an extension author you know whether that's the
+        // case or not, so Chrome reports this as an error. However, in order to
+        // polyfill the Promise version
+        // (`browser.runtime.sendMessage("some_message").then(response => {
+        // console.log(response) })`) we have to *always* use the callback,
+        // because the polyfill can't know if the other end will send a response
+        // or not. This makes the error irrelevant. Ignore it and resolve with
+        // nothing instead. See:
+        // https://github.com/mozilla/webextension-polyfill/issues/130
+        if (chrome.runtime.lastError.message === CHROME_SEND_MESSAGE_CALLBACK_NO_RESPONSE_MESSAGE) {
+          resolve();
+        } else {
+          reject(chrome.runtime.lastError);
+        }
       } else if (reply && reply.__mozWebExtensionPolyfillReject__) {
         // Convert back the JSON representation of the error into
         // an Error instance.
--- a/test/test-runtime-onMessage.js
+++ b/test/test-runtime-onMessage.js
@@ -239,4 +239,24 @@ describe("browser-polyfill", () => {
       });
     });
   });
+
+  it("ignores a certain chrome.runtime.lastError", () => {
+    const fakeChrome = {
+      runtime: {
+        lastError: {
+          message: "The message port closed before a response was received.",
+        },
+        sendMessage: sinon.stub(),
+      },
+    };
+
+    return setupTestDOMWindow(fakeChrome).then(window => {
+      fakeChrome.runtime.sendMessage
+        .onFirstCall().callsArgWith(1, ["res1", "res2"]);
+
+      return window.browser.runtime.sendMessage("some_message").then(
+        (...args) => deepEqual(args, [undefined], "The error was ignored")
+      );
+    });
+  });
 });


rpl:
@lydell Thanks for reporting and looking into this tricky issue, I'm a bit "on the fence" about silencing the errors (because it may theoretically make harder to find real issues that may trigger the same error message, and also on the other side if the error message changes we will not notice it and we will start to produce noise again), but I agree that these errors are going to be confusing given that the developers are not going to be in charge of passing the callbacks explicitly.

I'm going to give a try to your test extension asap, so that I can form a more precise opinion on the issue (and it may be my initial concerns are not really motivated).

In the meantime, as an initial feedback on the change:

- the inline comment is a bit too big ;-), I agree that it is a workaround that definitely requires some details but I would prefer a single (or no more than 2 lines) of inline comment and then the link to the issue #130 can easily provide any additional details in the issue description

- if "making the error silent" is our only option and we are going to proceed with this strategy, it could be worth to still produce some logs for the silenced errors when running in some kind of "dev mode" (e.g.   y checking if a certain global is defined, something like `typeof WEBEXT_POLYFILL_DEBUG !== "undefined"` or another similar kind of runtime check, or by building a separated browser-polyfill.devmode.js file), so that we can document for the addon developer how they can see the errors that may have been silenced (if they need to).

Rob--W:
@rpl I wouldn't worry about false positives. This error message is only generated when the `onMessage` handler is not invoked. In Chrome, errors (`chrome.runtime.lastError`) in `runtime.sendMessage`'s callback are silently swallowed, so it is better to ignore known errors too.

In Chrome, `runtime.sendMessage` is internally implemented as `runtime.connect` + `port.postMessage` at the sender's end, and `port.postMessage` + `port.disconnect` at the recipient (if they choose to respond). If there is no explicit response in the `onMessage` handler, the port is closed with the message `The message port closed before a response was received.`.

Without `--enable-features=NativeCrxBindings` : https://chromium.googlesource.com/chromium/src/+/a647c750dfb95d621807607c75939b8286c7acae/extensions/renderer/resources/messaging.js#437
With `--enable-features=NativeCrxBindings` : https://chromium.googlesource.com/chromium/src/+/a647c750dfb95d621807607c75939b8286c7acae/extensions/renderer/one_time_message_handler.cc#425

src/browser-polyfill.js
```diff
@@ -446,7 +448,23 @@ if (typeof browser === "undefined") {
 
     const wrappedSendMessageCallback = ({reject, resolve}, reply) => {
       if (chrome.runtime.lastError) {
-        reject(chrome.runtime.lastError);
+        // Passing a callback to `chrome.runtime.sendMessage`
```
rpl: Let's trim this inline comment a bit, and we should actually mention that this is needed to match the behavior on Firefox, e.g. something like the following could be enough:

```js
// Detect when none of the listener replied to the sendMessage call and resolve
// the promise to undefined as in Firefox.
// See https://github.com/mozilla/webextension-polyfill/issues/130        
```

test/test-runtime-onMessage.js
```diff
@@ -239,4 +239,24 @@ describe("browser-polyfill", () => {
       });
     });
   });
+
+  it("ignores a certain chrome.runtime.lastError", () => {
```
rpl: How about `it("resolves to undefined when no listeners reply"`?

test/test-runtime-onMessage.js
```diff
@@ -239,4 +239,24 @@ describe("browser-polyfill", () => {
...
+    const fakeChrome = {
+      runtime: {
+        lastError: {
+          message: "The message port closed before a response was received.",
```
rpl: let's add a 'single line' inline comment which mention that this is the error defined as CHROME_SEND_MESSAGE_CALLBACK_NO_RESPONSE_MESSAGE in the polyfill internals.

test/test-runtime-onMessage.js
```diff
@@ -239,4 +239,24 @@ describe("browser-polyfill", () => {
...
+        },
+        sendMessage: sinon.stub(),
+      },
+    };
+
+    return setupTestDOMWindow(fakeChrome).then(window => {
+      fakeChrome.runtime.sendMessage
```
rpl: Nit, do you mind to move this out of this callback? right after `const fakeChrome = {...};` at line 252.

test/test-runtime-onMessage.js
```diff
@@ -239,4 +239,24 @@ describe("browser-polyfill", () => {
...
+        .onFirstCall().callsArgWith(1, ["res1", "res2"]);
+
+      return window.browser.runtime.sendMessage("some_message").then(
+        (...args) => deepEqual(args, [undefined], "The error was ignored")
```
rpl: Nit, some small tweak to the assertions:
```js
const promise = window.browser.runtime.sendMessage("some_message");
ok(fakeChrome.runtime.sendMessage.calledOnce, "sendMessage has been called once");
return promise.then(reply => {
  deepEqual(reply, undefined, "sendMessage promise should be resolved to undefined");
});
```

PR #44, opened 2017-07-14
add more APIs to api-metadata

romantaraban:
Adding browser.permissions and browser.privacy.services.passwordSavingEnabled (https://bugzilla.mozilla.org/show_bug.cgi?id=1361364)

Issue #81, opened 2017-12-30
browser.tabs.getCurrent() resolves to empty array instead of undefined


dessant:
According to the docs browser.tabs.getCurrent() should return a promise which resolves to undefined when called from a context that has no browser tab. The API works as expected on Firefox, but resolves to an empty array on Chrome.

```js
browser.tabs.getCurrent().then(function(tab) {
    console.log(1, tab);
  })
chrome.tabs.getCurrent(function(tab) {
  console.log(2, tab);
});
```

Firefox background or popup script:
```js
1 undefined
2 undefined
```

Chrome background or popup script:
```js
1 Array(0)
2 undefined
```

andymckay:
Is this a bug in the polyfill library or a bug in Firefox? If its a bug in Firefox it should go into Bugzilla rather than here: https://bugzilla.mozilla.org/enter_bug.cgi?product=Toolkit&component=WebExtensions:%20Untriaged

dessant:
Polyfill.

PR #133, opened 2018-06-19
feat: Support promises in privacy namespace

Rob--W:
This PR adds promise support for the common subset of methods in the `privacy` namespace that are supported by both Chrome and Firefox.
In the future, we can consider extending the promise support to all objects in the `privacy` namespace, even if they are only supported by one browser.

Note that CI will currently fail because of bug #81 (which will be fixed by #132).

Supersedes and closes #44.

### DIFF
--- a/test/fixtures/privacy-setting-extension/content.js
+++ b/test/fixtures/privacy-setting-extension/content.js
@@ -1,0 +1,48 @@
+test("privacy API should be unavailable in the content script", (t) => {
+  t.deepEqual(browser.privacy, undefined, "browser.privacy should not be available in a content script");
+});
+
+test("privacy API should support promises", async (t) => {
+  async function callSettingAPI(method, ...args) {
+    // Invokes: browser.privacy.services.passwordSavingEnabled[method](...args);
+    let {type, result} = await browser.runtime.sendMessage({method, args});
+    // In Chrome `undefined` values are serialized to `null`, so check the type
+    // as determined in the background page.
+    return type === "undefined" ? undefined : result;
+  }
+
+  let res = await callSettingAPI("get", {});
+  t.deepEqual(res, {
+    levelOfControl: "controllable_by_this_extension",
+    // Enabled by default in Chrome; disabled by default in Firefox.
+    value: !navigator.userAgent.includes("Firefox/"),
+  }, "passwordSavingEnabled.get() resolves to the initial value");
+
+  const defaultValue = res.value;
+  const newValue = !defaultValue;
+  res = await callSettingAPI("set", {value: newValue});
+  if (navigator.userAgent.includes("Firefox/")) {
+    t.equal(res, true, "passwordSavingEnabled.set() resolves to true");
+  } else {
+    t.equal(res, undefined, "passwordSavingEnabled.set() resolves to a void value");
+  }
+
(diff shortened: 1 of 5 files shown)


test/fixtures/privacy-setting-extension/content.js
```diff
@@ -0,0 +1,48 @@
...
+    value: !navigator.userAgent.includes("Firefox/"),
+  }, "passwordSavingEnabled.get() resolves to the initial value");
+
+  const defaultValue = res.value;
+  const newValue = !defaultValue;
+  res = await callSettingAPI("set", {value: newValue});
+  if (navigator.userAgent.includes("Firefox/")) {
+    t.equal(res, true, "passwordSavingEnabled.set() resolves to true");
+  } else {
+    t.equal(res, undefined, "passwordSavingEnabled.set() resolves to a void value");
```
rpl: note to self: this different behavior of `set` and `clear` seems something that we should list in the known incompatibilities in the updated README (iirc on Firefox `set` and `clear` returns true if the setting has been changes, false otherwise, on the contrary Chrome doesn't pass any parameter to the callback that is called once the operation has been completed).

Rob--W:
Done: ```diff
diff --git a/test/fixtures/privacy-setting-extension/background.js b/test/fixtures/privacy-setting-extension/background.js
index 37aab05..ef7bf14 100644
--- a/test/fixtures/privacy-setting-extension/background.js
+++ b/test/fixtures/privacy-setting-extension/background.js
@@ -1,10 +1,5 @@
-console.log(name, "background page loaded");
-
 browser.runtime.onMessage.addListener(async (msg) => {
-  console.log(name, "background received msg", {msg});
   let {method, args} = msg;
   let result = await browser.privacy.services.passwordSavingEnabled[method](...args);
   return {result, type: typeof result};
 });
-
-console.log(name, "background page ready to receive a content script message...");
diff --git a/test/fixtures/privacy-setting-extension/content.js b/test/fixtures/privacy-setting-extension/content.js
index 5a8ce46..78cc4ef 100644
--- a/test/fixtures/privacy-setting-extension/content.js
+++ b/test/fixtures/privacy-setting-extension/content.js
@@ -1,13 +1,13 @@
 test("privacy API should be unavailable in the content script", (t) => {
-  t.deepEqual(browser.privacy, undefined, "browser.privacy is not available in a content script");
+  t.deepEqual(browser.privacy, undefined, "browser.privacy should not be available in a content script");
 });
 
 test("privacy API should support promises", async (t) => {
   async function callSettingAPI(method, ...args) {
     // Invokes: browser.privacy.services.passwordSavingEnabled[method](...args);
     let {type, result} = await browser.runtime.sendMessage({method, args});
-    // In Chrome `undefined` values are serialized to `null`, so use the
-    // pre-determined type instead.
+    // In Chrome `undefined` values are serialized to `null`, so check the type
+    // as determined in the background page.
     return type === "undefined" ? undefined : result;
   }
 
```

test/fixtures/privacy-setting-extension/content.js
Rob--W: This seems like something that should be documented on MDN: https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/types/BrowserSetting

Issue #135, opened 2018-06-19
The build passes despite the tests failing in Chrome


ExE-Boss:
See #133, where the tests fail in Chrome but pass in Firefox, which results in the build being marked as passing (✔) instead of failing (❌):
https://travis-ci.org/mozilla/webextension-polyfill/builds/394295666#L2377

PR #136, opened 2018-06-19
Fix the `\n` character preceding `echo` output in shell scripts

ExE-Boss:
This fixes the `\n` character sequence preceding all `echo` output in the shell scripts:
eg.: https://travis-ci.org/mozilla/webextension-polyfill/builds/394310210#L2181

### DIFF
--- a/test/run-browsers-smoketests.sh
+++ b/test/run-browsers-smoketests.sh
@@ -1,5 +1,7 @@
-echo "\nTest webextension-polyfill on real browsers"
-echo "============================================"
+#!/bin/bash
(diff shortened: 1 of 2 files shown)


ExE-Boss:
Output without this:
https://travis-ci.org/mozilla/webextension-polyfill/builds/394310210#L2181

Output with this:
https://travis-ci.org/mozilla/webextension-polyfill/builds/394325525#L2190

Rob--W:
Thanks. I've merged it so that you don't have to rebase when I fix #135.

PR #137, opened 2018-06-19
chore: Stop swallowing errors in .sh tests

### LINKED ISSUE #135: The build passes despite the tests failing in Chrome

Rob--W:
See the commit description.

Fixes #135

### DIFF
--- a/test/run-browsers-smoketests.sh
+++ b/test/run-browsers-smoketests.sh
@@ -1,4 +1,6 @@
 #!/bin/bash
+set -eo pipefail
+
 echo ""
 echo "Test webextension-polyfill on real browsers"
 echo "==========================================="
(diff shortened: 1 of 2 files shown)


ExE-Boss:
This will unfortunately cause the Firefox tests to be skipped if Chrome fails.

I propose to instead use a temporary `RESULT` variable to store non‑zero exit values to.

Rob--W:
> `set -e` will unfortunately cause the Firefox tests to be skipped if the Chrome tests fail.

That is no different from how tests are currently run. E.g. if linting fails, then the code won't be built and run. Any PRs with failing CI won't be merged, so if the Chrome tests fail already, then there is no real need to also run the Firefox tests.

PR #140, opened 2018-07-02
fix: sendMessage promise should resolve to undefined when no listeners reply

### LINKED ISSUE #130: Chrome: "The message port closed before a response was received."

rpl:
This PR contains the changes from #131 with the addition of some small tweaks (from the last review on #131) and an additional integration test (to check that the behavior of the polyfill in Chrome is the same of the one on Firefox with the native promised APIs).

Fixes #130

### DIFF
--- a/test/fixtures/runtime-messaging-extension/content.js
+++ b/test/fixtures/runtime-messaging-extension/content.js
@@ -75,3 +75,8 @@ test("sendMessage with listener callback throws", async (t) => {
     t.equal(err.message, "listener throws", "Got an error with the expected message");
   }
 });
+
+test("sendMessage and no listener answers", async (t) => {
+  const reply = await browser.runtime.sendMessage("test - sendMessage and no listener answers");
+  t.equal(reply, undefined, "Got undefined reply as expected");
+});
--- a/test/test-runtime-onMessage.js
+++ b/test/test-runtime-onMessage.js
@@ -238,5 +238,30 @@ describe("browser-polyfill", () => {
         });
       });
     });
+
+    it("resolves to undefined when no listeners reply", () => {
+      const fakeChrome = {
+        runtime: {
+          // This error message is defined as CHROME_SEND_MESSAGE_CALLBACK_NO_RESPONSE_MESSAGE
+          // in the polyfill sources and it is used to recognize when Chrome has detected that
+          // none of the listeners replied.
+          lastError: {
+            message: "The message port closed before a response was received.",
+          },
+          sendMessage: sinon.stub(),
+        },
+      };
+
+      fakeChrome.runtime.sendMessage.onFirstCall().callsArgWith(1, [undefined]);
+
+      return setupTestDOMWindow(fakeChrome).then(window => {
+        const promise = window.browser.runtime.sendMessage("some_message");
+        ok(fakeChrome.runtime.sendMessage.calledOnce, "sendMessage has been called once");
+
+        return promise.then(reply => {
+          deepEqual(reply, undefined, "sendMessage promise should be resolved to undefined");
+        });
+      });
+    });
   });
 });
--- a/src/browser-polyfill.js
+++ b/src/browser-polyfill.js
@@ -7,6 +7,7 @@
 "use strict";
 
 if (typeof browser === "undefined") {
+  const CHROME_SEND_MESSAGE_CALLBACK_NO_RESPONSE_MESSAGE = "The message port closed before a response was received.";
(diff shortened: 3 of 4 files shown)


test/fixtures/runtime-messaging-extension/content.js
```diff
@@ -75,3 +75,8 @@ test("sendMessage with listener callback throws", async (t) => {
     t.equal(err.message, "listener throws", "Got an error with the expected message");
   }
 });
+
+test("sendMessage and no listener answers", async (t) => {
+  const reply = await browser.runtime.sendMessage("test - sendMessage and no listener answers");
+  t.equal(reply, undefined, "Got undefined as a reply a expected");
```
lydell: Typo: a<strong>s</strong> expected

test/test-runtime-onMessage.js
```diff
@@ -238,5 +238,30 @@ describe("browser-polyfill", () => {
...
+          // in the polyfill sources and it is used to recognize when Chrome has detected that
+          // none of the listeners replied.
+          lastError: {
+            message: "The message port closed before a response was received.",
+          },
+          sendMessage: sinon.stub(),
+        },
+      };
+
+      fakeChrome.runtime.sendMessage.onFirstCall().callsArgWith(1, ["res1", "res2"]);
```
Rob--W: Nit: Use `.callsArg(1)` or `callsArgWith(1, [undefined]);` .
In reality the `sendMessage` callback won't ever be invoked with two arguments.

PR #141, opened 2018-07-03
Minify the API metadata during build

ExE-Boss:
This makes the build step minify imported JSON files into JavaScript objects without altering the semantic meaning.

See https://github.com/mozilla/webextension-polyfill/pull/48#issuecomment-401921532 and https://github.com/mozilla/webextension-polyfill/pull/48#issuecomment-402180153.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -3,6 +3,7 @@
 "use strict";
 
 /* eslint-env commonjs */
+const path = require("path");
 
 const LICENSE = `/* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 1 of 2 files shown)


Rob--W:
Apparently, during minification, unnecessary quotes are already being stripped:

- https://unpkg.com/webextension-polyfill@0.2.1/dist/browser-polyfill.js
- https://unpkg.com/webextension-polyfill@0.2.1/dist/browser-polyfill.min.js

Were you aware of that? What's the objective of this PR?

ExE-Boss:
Actually, the `apiMetadata` object in https://unpkg.com/webextension-polyfill@0.2.1/dist/browser-polyfill.js doesn’t have the unnecessary quotes stripped and takes ~800 lines.

The objective of this PR is to have the `apiMetadata` object take a single line and be as small as possible..

james-cnz:
I was wondering about reducing the metadata to one method per line directly in the source, like this:
https://github.com/james-cnz/webextension-polyfill/blob/squash_metadata_method_per_line/api-metadata.json
This saves ~26%, and, I think, actually improves readability, without requiring any build-time processing.

I guess it would be pretty easy and unproblematic to strip all quotes during the build (are all quotes unnecessary), like this:
https://github.com/james-cnz/webextension-polyfill/blob/squash_metadata_method_per_line_no_quotes/api-metadata.json
This saves a further ~9%, without making the code less readable, or breaking any JavaScript style guides, I think?

Removing all line breaks, but leaving spaces where (I think they are) required for style (is this the intent of the PR), saves a further ~7%, but at a cost to readability.  I'm not sure this worthwhile for the unminified target?

PR #151, opened 2018-07-19
chore: Use correct node_modules/ in shell script

Rob--W:
Use an absolute path so that the shell scripts use the expected node_modules directory (at the repository root) instead of whatever "node_modules" directory is in the current directory.

### DIFF
--- a/test/run-browsers-smoketests.sh
+++ b/test/run-browsers-smoketests.sh
@@ -1,11 +1,12 @@
 #!/bin/bash
 set -eo pipefail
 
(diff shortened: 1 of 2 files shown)


rpl:
Closing, as this wasn't such a big deal (especially given that those scripts were usually executed as npm scripts and npm takes care of making the node_modules binaries available in the PATH env var while running npm scripts) and because this PR can't be applied anymore (as these are not shell scripts anymore, they have been rewritten into nodejs scripts as part of #167).

PR #153, opened 2018-07-21
fix: Prevent a webpage document element to be detected as the Extension API object when running in content scripts

rpl:
This PR applies some changes to the webpage used by the integration test suite, to reproduce the same issue that is described in this [comment on issue #68](https://github.com/mozilla/webextension-polyfill/issues/68#issuecomment-406785652) and on https://crbug.com/865881

And the additional check `browser instanceof window.Node` in the polyfill sources, to detect this scenario and prevent the DOM node to be detected as an existing Extension API object (and turn the polyfill into a no-op as when it runs on Firefox where the browser API object is natively available).

### DIFF
--- a/src/browser-polyfill.js
+++ b/src/browser-polyfill.js
@@ -6,7 +6,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 "use strict";
 
-if (typeof browser === "undefined" || browser instanceof window.Node) {
+if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.prototype) {
   const CHROME_SEND_MESSAGE_CALLBACK_NO_RESPONSE_MESSAGE = "The message port closed before a response was received.";
   const SEND_RESPONSE_DEPRECATION_WARNING = `
       Returning a Promise is the preferred way to send a reply from an
--- a/src/browser-polyfill.js
+++ b/src/browser-polyfill.js
@@ -6,7 +6,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 "use strict";
 
-if (typeof browser === "undefined") {
+if (typeof browser === "undefined" || browser instanceof window.Node) {
   const CHROME_SEND_MESSAGE_CALLBACK_NO_RESPONSE_MESSAGE = "The message port closed before a response was received.";
   const SEND_RESPONSE_DEPRECATION_WARNING = `
       Returning a Promise is the preferred way to send a reply from an
--- a/test/fixtures/detect-existing-browser-api-object/background.js
+++ b/test/fixtures/detect-existing-browser-api-object/background.js
@@ -1,0 +1,15 @@
+/* global originalAPIObjects */
+
+// Register an additional message handler that always reply after
+// a small latency time.
+browser.runtime.onMessage.addListener(async (msg, sender) => {
+  if (msg !== "test-api-object-in-background-page") {
+    return false;
+  }
+
+  return Promise.resolve({
+    browserIsDefined: !!browser,
+    chromeIsDefined: !!chrome,
+    browserIsUnchanged: browser === originalAPIObjects.browser,
+  });
+});
--- a/test/fixtures/detect-existing-browser-api-object/background.js
+++ b/test/fixtures/detect-existing-browser-api-object/background.js
@@ -1,15 +1,14 @@
 /* global originalAPIObjects */
 
-// Register an additional message handler that always reply after
-// a small latency time.
-browser.runtime.onMessage.addListener(async (msg, sender) => {
+browser.runtime.onMessage.addListener(async (msg, sender, sendResponse) => {
   if (msg !== "test-api-object-in-background-page") {
-    return false;
+    throw new Error(`Unexpected message received: ${msg}`);
   }
 
-  return Promise.resolve({
+  return {
     browserIsDefined: !!browser,
     chromeIsDefined: !!chrome,
     browserIsUnchanged: browser === originalAPIObjects.browser,
-  });
+    windowBrowserIsUnchanged: window.browser === originalAPIObjects.browser,
+  };
 });
--- a/test/fixtures/detect-existing-browser-api-object/copy-original-api-objects.js
+++ b/test/fixtures/detect-existing-browser-api-object/copy-original-api-objects.js
@@ -1,0 +1,2 @@
+// Store a copy of the references to the original API objects.
+const originalAPIObjects = {browser, chrome}; // eslint-disable-line no-unused-vars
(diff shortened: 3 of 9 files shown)


src/browser-polyfill.js
```diff
@@ -6,7 +6,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 "use strict";
 
-if (typeof browser === "undefined") {
+if (typeof browser === "undefined" || browser instanceof window.Node) {
```
tophf: Shouldn't it be `window.browser instanceof window.Node` to avoid an exception in strict mode?

Rob--W: Change to:

```javascript
if (typeof browser !== "object" || browser && Object.getPrototypeOf(browser) !== Object.prototype) {`
```

or (if you prefer shorter code or don't want to rely on `Object.getPrototypeOf`):

```javascript
if (typeof browser === "undefined" || browser && browser.__proto__ !== Object.prototype) {
```

or (if you don't want to rely on `Object`):

```javascript
if (typeof browser === "undefined" || browser && browser.__proto__ !== ({}).__proto__) {
```

Why? Well:
1. Using `instanceof window.Node` won't rule out tampering. You'll see that in this case, `x` is a `HTMLCollection`: `data:text/html,<a id=x href>a</a><a id=x href>b</a><script>console.log(x)</script>`
2. If someone has happened to define a primitive value as "browser", then we still want to overwrite it. (this is an accidental side-effect of my chosen implementation)

See also https://www.w3.org/TR/html5/browsers.html#named-access-on-the-window-object

test/fixtures/index.html
```diff
@@ -6,5 +6,12 @@
   </head>
   <body>
     <h1>Browser Polyfill Test Page</h1>
+
+    <!--
+         The following two DOM elements ensure that the polyfill doesn't detect the
+         globals defined for these DOM element ids as the Extensions API objects.
+    -->
+    <div id="browser"></div>
```
Rob--W: Add another `<div id="browser"></div>` to demonstrate the issue that I mentioned above.

rpl:
@tophf have you deleted Rob comment by mistake?

Follows @Rob--W deleted comment for reference

-----

In src/browser-polyfill.js:

```diff
> @@ -6,7 +6,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 "use strict";
 
-if (typeof browser === "undefined") {
+if (typeof browser === "undefined" || browser instanceof window.Node) {
```

Change to:

```js
if (typeof browser !== "object" || browser && Object.getPrototypeOf(browser) !== Object.prototype) {`
```
or (if you prefer shorter code or don't want to rely on Object.getPrototypeOf):

```js
if (typeof browser === "undefined" || browser && browser.__proto__ !== Object.prototype) {
```

or (if you don't want to rely on Object):

```js
if (typeof browser === "undefined" || browser && browser.__proto__ !== ({}).__proto__) {
```

Why? Well:
- Using instanceof window.Node won't rule out tampering. You'll see that in this case, x is a HTMLCollection: data:text/html,<a id=x href>a</a><a id=x href>b</a><script>console.log(x)</script>
- If someone has happened to define a primitive value as "browser", then we still want to overwrite it. (this is an accidental side-effect of my chosen implementation)

See also https://www.w3.org/TR/html5/browsers.html#named-access-on-the-window-object

test/fixtures/index.html
```diff
@@ -6,5 +6,12 @@
   </head>
   <body>
     <h1>Browser Polyfill Test Page</h1>
+
+    <!--
+         The following two DOM elements ensure that the polyfill doesn't detect the
```
Rob--W: test

rpl: test2

tophf:
>@tophf have you deleted Rob comment by mistake?

I don't have any rights in this repo so I couldn't have deleted anyone else's comments directly. I guess github created a "thread" for my comment and Rob commented in that thread. Then I realized my comment is wrong and deleted it, which made github delete the entire "thread".

rpl:
> Do we already have an integration test that verifies that the polyfill does not overwrite the browser global in Firefox (content scripts and extension/background page)?

@Rob--W 

We have a unit test (which simulates the behaviors in an extension page/background page):
https://github.com/mozilla/webextension-polyfill/blob/450eee59a32afeed8b286d02f7446b8fbaacd1ce/test/test-browser-global.js#L15-L27

And an integration test which tests the behavior in a content script context on Chrome and Firefox:
https://github.com/mozilla/webextension-polyfill/blob/450eee59a32afeed8b286d02f7446b8fbaacd1ce/test/fixtures/detect-browser-api-object-in-content-script/content.js#L1-L18

Rob--W:
> And an integration test which tests the behavior in a content script context on Chrome and Firefox:
https://github.com/mozilla/webextension-polyfill/blob/450eee59a32afeed8b286d02f7446b8fbaacd1ce/test/fixtures/detect-browser-api-object-in-content-script/content.js#L1-L18

The way that we inject APIs in Firefox are different for extension pages and content scripts, so a test for the background page seems desired.

`browser` being identical to `chrome` in a content script seems to be a small bug to me, because it adds promise APIs to the `chrome` namespace in content scripts: https://bugzilla.mozilla.org/show_bug.cgi?id=1296896
In extension pages, `chrome` and `browser` are already different. We just need to make sure that the polyfill does not replace the `browser` in this case.

test/fixtures/detect-existing-browser-api-object/background.js
```diff
@@ -0,0 +1,15 @@
+/* global originalAPIObjects */
+
+// Register an additional message handler that always reply after
+// a small latency time.
+browser.runtime.onMessage.addListener(async (msg, sender) => {
+  if (msg !== "test-api-object-in-background-page") {
+    return false;
```
Rob--W: Why `return false;`? If you don't want the handler to respond, jus return void (`return;`).
Or make this a non-async function and call `sendResponse({ ... });` in the end.

test/fixtures/detect-existing-browser-api-object/copy-original-api-objects.js
```diff
@@ -0,0 +1,2 @@
+// Store a copy of the references to the original API objects.
+const originalAPIObjects = {browser, chrome}; // eslint-disable-line no-unused-vars
```
Rob--W: Add a comment that in browsers without a `browser` API, this will point to the `<div id="browser">` in the page.

test/fixtures/detect-existing-browser-api-object/background.js
```diff
@@ -0,0 +1,15 @@
...
+  }
+
+  return Promise.resolve({
+    browserIsDefined: !!browser,
+    chromeIsDefined: !!chrome,
+    browserIsUnchanged: browser === originalAPIObjects.browser,
```
Rob--W: Let's check `window.browser` too?

PR #154, opened 2018-07-27
Support Microsoft Edge

PoziWorld:
Working example: https://github.com/PoziWorld/Scroll-to-Top-Button-Extension/releases/tag/v7.0.0


rpl:
Thanks @PoziWorld but we already have another pull request about supporting Edge, we have deferred further work on it while we were release the new 0.3.0 version, but #114 should be near enough to the goal.

Closing this one in favor of #114.

PR #163, opened 2018-11-26
test(integration): Add support for Edge to the integration test suite

rpl:
This PR applies the changes needed to be able to run the entire integration test suite on Edge. 

It is true that most windows CI services can't currently run Edge because it can't be installed on Windows server systems, which is quite unfortunate, but to be able to realistically "support the webextension-polyfill on Edge (even on a best-effort basis)"... the bare minimum is to at least ensure that all the existing tests can be run successfully on a regular Windows system).

And so this PR is propaedeutic to be able to complete and merge the PR #114, which is meant to allow the polyfill to support Edge.

In this PR that are also the following small changes: 

- on the polyfill sources, to fix another small issue when the polyfill implementation is running on Edge (which has not been caught yet in #114, I noticed it while running the entire test suite on Edge)

- define new npm scripts (to make it easier to run the integration tests on a single browser (e.g. "npm run test-integration:edge" will run the tests on just Edge)

- rewritten one bash script into a nodejs script (to make it easier to run the integration tests across the supported browsers on windows, without the need to have a bash compatible shell available on the windows systems

- fixed minor incompatibilities of the test extensions' manifest.json files (e.g. Edge will not run the extension if the manifest doesn't have an author property, or the persistent property is missing in the background page manifest section, and the content scripts are not running on the test server listening on localhost without changing the origin permission to "http://*/*")

- Allow to skip an entire test extension (e.g. on some browsers)

- Make it a bit easier to investigate issues in the test extensions (by temporarily asking the integration test suite to keep the controlled browser opened after the test is completed)


test/integration/setup.js
```diff
@@ -71,9 +71,42 @@ const launchBrowser = async (launchOptions) => {
...
 
     await driver.execute(command);
+  } else if (browser === "edge") {
+    const edge = require("selenium-webdriver/edge");
+    const edgeLocalStatePath = path.join(process.env.LOCALAPPDATA,
+                                         "Packages", "Microsoft.MicrosoftEdge_8wekyb3d8bbwe", "LocalState");
+    const destExtensionDir = path.join(edgeLocalStatePath, path.basename(extensionPath) + "-" + Date.now());
+
+    const shell = require("shelljs");
+    console.log("Copying test extension from", extensionPath, "to", destExtensionDir);
```
Rob--W: Use template literals.

test/integration/setup.js
```diff
@@ -71,9 +71,42 @@ const launchBrowser = async (launchOptions) => {
...
+    const options = new edge.Options();
+    options.set("extensionPaths", [destExtensionDir]);
+
+    let edgeDriver = await new Builder()
+      .forBrowser("MicrosoftEdge")
+      .setEdgeOptions(options)
+      .setEdgeService(new edge.ServiceBuilder())
+      .build();
+    driver = Object.create(null);
+    Object.setPrototypeOf(driver, edgeDriver);
```
Rob--W: Why not: `driver = Object.create(edgeDriver);`

test/integration/setup.js
```diff
@@ -98,13 +131,16 @@ const createHTTPServer = async (path) => {
...
   try {
     const url = `http://localhost:${server.address().port}`;
     const userAgent = await driver.executeScript(() => window.navigator.userAgent);
 
-    t.pass(`Connected to browser: ${userAgent}"`);
-
+    t.pass(`Connected to ${browser} browser: ${userAgent}"`);
+    if (browser === "edge") {
+      // Edge seems to need a little latency to be able to start the background page.
+      await driver.sleep(500);
```
Rob--W: This is prone to time-outs. Is it feasible to notify the test runner when the background page is ready? E.g. by sending a request on start-up of the background page.

test/integration/setup.js
```diff
@@ -138,23 +174,40 @@ const bundleTapeStandalone = async (destDir) => {
 
   const stream = b.bundle();
   const onceStreamEnd = awaitStreamEnd(stream);
-  stream.pipe(fs.createWriteStream(bundleFileName));
+  const destFileStream = fs.createWriteStream(bundleFileName);
+  const onceWritten = new Promise(resolve => {
+    destFileStream.on("close", resolve);
```
Rob--W: If you're doing this anyway, consider a `destFileStream.on("error", reject);` too (in case somehow the destination file is non-writable, e.g. due to a permission issue).

test/integration/setup.js
```diff
@@ -169,16 +222,24 @@ const defineExtensionTests = ({description, extensions}) => {
...
-          runExtensionTest(tt, server, driver, extensionDirName),
+          runExtensionTest(tt, server, driver, extensionDirName, browser),
           new Promise((resolve, reject) => {
             timeout = setTimeout(() => reject(new Error(`test timeout after ${TEST_TIMEOUT}`)), TEST_TIMEOUT);
           }),
         ]);
       } finally {
         clearTimeout(timeout);
         if (driver) {
+          if (exit === false) {
```
Rob--W: Where is `exit` explicitly set to `false`? It doesn't seem to be used this PR.

Even if you do end up using `exit`, I think that the `server` cleanup logic below should run too (currently it does not due to the use of a never-resolving Promise.

By the construction of this logic, it looks like auxilary debugging code that helped during development, but shouldn't be checked in.

test/integration/setup.js
```diff
@@ -138,23 +174,40 @@ const bundleTapeStandalone = async (destDir) => {
...
       let server;
       let tempDir;
 
+      let browser = process.env.TEST_BROWSER_TYPE;
+
+      if (skip) {
+        if (skip === true) {
+          tt.skip("Test extension skipped");
+          return;
+        } else if (skip === browser || skip.includes(browser)) {
```
Rob--W: This adds three ways to disable a test: `true`, string, array of strings. Onle the string value is used (`skip: "edge"`). If there is no use for the other two, let's keep it simple for now and only use the string value.

Or remove this alltogether and skip the test in the description inside the test file itself.

test/fixtures/multiple-onmessage-listeners-extension/content.js
```diff
@@ -1,13 +1,14 @@
 test("Multiple runtime.onmessage listeners which resolve to undefined", async (t) => {
   const res = await browser.runtime.sendMessage("test-multiple-onmessage-listeners:resolve-to-undefined");
 
-  if (navigator.userAgent.includes("Firefox/")) {
-    t.deepEqual(res, undefined, "Got an undefined value as expected");
-  } else {
+  if (navigator.userAgent.includes("Chrome/") && !navigator.userAgent.includes("Edge/")) {
```
Rob--W: This is because Edge pretends to be Chrome, right?

The intent might be a bit clearer if you do UA sniffing in one location (e.g. test/fixtures/tape-standalone.js ) and then export the environment type to a variable (e.g. `TEST_BROWSER_NAME` or something like that). I'm using a variable instead of a property so that if you mistakenly use this global variable in a context where it is not defined, that a ReferenceError is triggered instead of getting the `undefined` value.

test/integration/setup.js
```diff
@@ -71,9 +71,42 @@ const launchBrowser = async (launchOptions) => {
...
+    const cleanup = async () => {
+      console.log("Removing temp extension dir", destExtensionDir);
+      shell.rm("-rf", destExtensionDir);
+    };
+    driver.close = async () => {
+      await edgeDriver.close();
+      await cleanup();
+    };
+    driver.quit = async () => {
+      // await new Promise(() => {});
```
Rob--W: Stray comment?

scripts/run-browsers-smoketests.js
```diff
@@ -0,0 +1,29 @@
...
+
+// set -eo pipefail
+
+console.log(`
+Test webextension-polyfill on real browsers
+===========================================
+`);
+
+if (process.platform == "win32" && process.env.EDGE_DRIVER_PATH) {
+  process.env.PATH += `;${process.env.EDGE_DRIVER_PATH}`;
```
Rob--W: Where is this environment variable documented? If it is only used as a convenience to set the location of the Edge driver, shouldn't this be in the test-integration implementation?

If the line is to be kept, let's move this line into the block with the same condition at the end of this file.

test/integration/setup.js
ExE-Boss:
```suggestion
    console.log(`Copying test extension from ${extensionPath} to ${destExtensionDir}`);
```

test/integration/setup.js
```diff
@@ -138,23 +174,40 @@ const bundleTapeStandalone = async (destDir) => {
 
   const stream = b.bundle();
   const onceStreamEnd = awaitStreamEnd(stream);
-  stream.pipe(fs.createWriteStream(bundleFileName));
+  const destFileStream = fs.createWriteStream(bundleFileName);
+  const onceWritten = new Promise(resolve => {
```
ExE-Boss:
```suggestion
  const onceWritten = new Promise((resolve, reject) => {
    destFileStream.on("error", reject);
```

rpl:
Closing as wontfix. See https://github.com/mozilla/webextension-polyfill/pull/114#issuecomment-451923835 for a rationale.

PR #166, opened 2019-01-08
feat: Add schema and type definition for `api‑metadata.json`

ExE-Boss:
This adds a&nbsp;JSON&nbsp;Schema and a&nbsp;TypeScript definition file for&nbsp;<code>api&#x2011;metadata.json</code>.

## Additional changes:
- Updates JSDoc comments to&nbsp;make use of&nbsp;the&nbsp;new&nbsp;TypeScript definition file.

## To&nbsp;do:
- [ ] Use `ajv` and <code>better&#x2011;ajv&#x2011;errors</code> to&nbsp;validate <code>api&#x2011;metadata.json</code> against <code>api&#x2011;metadata.schema.json</code>

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -429,7 +429,7 @@
     "sort-vars": 0,
 
     // Require a space immediately following the // in a line comment.
-    "spaced-comment": [2, "always"],
+    "spaced-comment": [2, "always", { "markers": ["/"] }],
 
     // Require "use strict" to be defined globally in the script.
     "strict": [0, "global"],
--- a/api-metadata.schema.json
+++ b/api-metadata.schema.json
@@ -1,0 +1,59 @@
+{
+  "definitions": {
+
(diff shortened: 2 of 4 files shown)


Rob--W:
The only relevant interface for public consumers is the exported `browser` API. Definitions for these extension APIs should already exist.

The internal API does not need to have such verbose (boilerplate) documentation, so we are going to reject that part.

The schema is potentially useful, but given the few properties that we support, a human-readable document would be more fitting (if any). The property names are self-explanatory (and if not, the implementation, commit history and/or code review would clear things up), so let's not bother with adding more text for now.

We are closing the issue for the above reasons; If I missed anything, feel free to comment.

ExE-Boss:
The type definition is&nbsp;there for&nbsp;intellisense in&nbsp;**VS&nbsp;Code** when developing the&nbsp;polyfill.

PR #167, opened 2019-01-08
 test(integration): Use node for running test scripts

ExE-Boss:
This adds @rpl’s changes from #163 that convert the&nbsp;bash scripts to node scripts, making testing easier on&nbsp;Windows (the&nbsp;platform I&nbsp;use for&nbsp;development).

review?(@Rob--W, @rpl)


Rob--W:
I haven't reviewed the PR in detail, but I approve of the motivation behind it.
What are the high-level differences between your PR and the original? I see that some of my requested changes (e.g. the pipefail one) has not been addressed in the first commit, but they have been addressed separately in the second commit.

ExE-Boss:
The primary difference is the removal of&nbsp;Edge&#x2011;specific code.

Then there’s the&nbsp;ESLint configuration improvements and better Windows compatibility.

scripts/run-browsers-smoketests.js
```diff
@@ -0,0 +1,24 @@
...
+console.log(`
+Test webextension-polyfill on real browsers
+===========================================`);
+
+// Enable headless mode (currently only used when running on Firefox
+// because Chrome doesn't currently support the extensions in headless mode)
+process.env.HEADLESS = 1;
+
+console.log(`
+Runing smoketests on Chrome`);
```
rpl: Typo: Runing -> Running

(Besides that, personally I would prefer a \n instead of a template string that starts with a new line, and same comment also for the similar template strings in run-module-bundlers-smoketests.js).

scripts/run-browsers-smoketests.js
```diff
@@ -0,0 +1,24 @@
...
+process.env.TEST_BROWSER_TYPE = "chrome";
+result = shell.exec("npm run test-integration:chrome").code || result;
```
rpl: How about early exit with `process.exit(-1)` (or with `process.exit(execResult.code)`) on the first `shell.exec` that fails?

to be honest, shelljs has a `shell.set` method which could allow us to solve this scenarios as it was solved in the original bash script:
- https://github.com/shelljs/shelljs#setoptions

I never used it, but it seems to be worth a try.

@ExE-Boss do you mind to give a try to `shell.set`?

(same comment on run-module-bundlers-smoketests.js)

scripts/run-module-bundlers-smoketests.js
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env node
+const shell = require("shelljs");
+
+let result = 0;
+
+console.log(`
+Test webextension-polyfill bundled with webpack
+===============================================`);
+
+shell.exec("webpack --mode production --entry ./test/fixtures/bundle-entrypoint.js --output /tmp/webpack-bundle.js");
```
rpl: we should early exit if webpack (or browserify) commands fails to generate the bundle.

Let's give a try to shelljs' set (as mentioned in the previous comment), if that doesn't work, let's explicitly check the code of these shell.exec results.

scripts/run-module-bundlers-smoketests.js
```diff
@@ -0,0 +1,22 @@
...
+process.env.TEST_BUNDLED_POLYFILL = "/tmp/webpack-bundle.js";
```
rpl: Nit, in my opinion setting the process.env.TEST_BUNDLED_POLYFILL and then the two shell.exec calls may be a bit more readable, I mean something like:

```
process.env.TEST_BUNDLED_POLYFILL = "...";
shelljs.exec("webpack ...");
shelljs.exec("npm run test"):
```

test/integration/setup.js
```diff
@@ -138,23 +138,46 @@ const bundleTapeStandalone = async (destDir) => {
...
   for (const extensionDirName of extensions) {
     test(`${description} (test extension: ${extensionDirName})`, async (tt) => {
       let timeout;
       let driver;
       let server;
       let tempDir;
 
+      const browser = process.env.TEST_BROWSER_TYPE;
+
+      if (skip) {
```
rpl: By removing from #163 the pieces related to Edge, there is no use of the skip property, and so we may choose to defer it or remove it as @Rob--W suggested in #163.

src/browser-polyfill.js
```diff
@@ -509,7 +509,7 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
 
   // The build process adds a UMD wrapper around this file, which makes the
   // `module` variable available.
-  module.exports = wrapAPIs(chrome); // eslint-disable-line no-undef
```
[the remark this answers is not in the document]
ExE-Boss: Well, the&nbps;ESLint CommonJS environment only allows `require` and&nbsp;`module.exports`:

https://github.com/sindresorhus/globals/blob/06fb76415e4c1f664c8e0a1185f800956b59fd45/globals.json#L1029-L1034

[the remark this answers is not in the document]
rpl: @ExE-Boss I know what the commonjs eslint environment is explicitly allowing, and that's exactly what I mean in https://github.com/mozilla/webextension-polyfill/pull/167#discussion_r246878903:

- I prefer that we prevent `require` to be used in this UMD module, because the UMD module is supposed to be self-contained (and this is also a reason why this line https://github.com/mozilla/webextension-polyfill/blob/master/src/browser-polyfill.js#L22 is not a `require` statement)

- I prefer to be sure that eslint is preventing `module.exports` to be used anywhere else, because the UMD module is supposed to export only that single property and I want to be sure that no other part of the polyfill ever tries to exports anything else.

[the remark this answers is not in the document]
Rob--W: I agree with @rpl's assessment above. Currently there are three variables in the line, and the intention is to apply the var only to `module.exports`. If you really like to minimize the usage of the eslint rule, then you could use:

```javascript
module.exports = // eslint-disable-line no-undef
  wrapAPIs(chrome);
```

I think that leaving the code untouched (before this PR) is also fine though.

scripts/run-browsers-smoketests.js
ExE-Boss: As&nbsp;I added to&nbsp;the&nbsp;documentation of&nbsp;the&nbsp;`result` variable:

> This is to make sure that even if the tests fail on Chrome, the tests still run on Firefox, so that it can be seen whether Firefox broke too or is unaffected.

src/browser-polyfill.js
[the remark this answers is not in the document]
ExE-Boss: I decided to&nbsp;declare <code>module:&nbsp;true</code> in&nbsp;the&nbsp;`.eslintrc` for&nbsp;the&nbsp;`src` directory.

Issue #168, opened 2019-01-10
browser.devtools.inspectedWindow.eval() resolves with incorrect promise signature


keithclark:
When calling `browser.devtools.inspectedWindow.eval()` in Chrome (I'm using 71), the resulting promise appears to resolve with an incorrect signature. According to [MDN](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/devtools.inspectedWindow/eval#Return_value), the resulting promise should be an array containing two elements:

> A Promise that will be fulfilled with an array containing two elements.
>
>If no error occurred, element 0 will contain the result of evaluating the expression, and element 1 will be undefined.
>
>If an error occurred, element 0 will be undefined, and element 1 will contain an object giving details about the error. Two different sorts of errors are distinguished

However, running the following script

```js
let statement = `Array.from(document.querySelectorAll('*')).map(e=>e.tagName)`;
browser.devtools.inspectedWindow.eval(statement).then(console.log);
```

Returns

```js
["HTML", "HEAD", "META", ...]
```

I was expecting

```js
[
  ["HTML", "HEAD", "META", ...],
  undefined
]
```

If I call the same function with a statement that will throw an error, the promise resolves with the correct error signature:

```js
let statement = `%%%`;
browser.devtools.inspectedWindow.eval(statement).then(console.log);
```

Returns:

```js
[
  undefined,
  {
    isException: true,
    value: "SyntaxError: Unexpected token %"
  }
]
```

Firefox behaves as expected.

Rob--W:
Chrome's devtools API implementation is not implemented using the default extension bindings. The implementation is here, and contrary to [Chrome's documentation](https://developer.chrome.com/extensions/devtools_inspectedWindow#property-eval-callback), there are not always two parameters passed to the callback: https://chromium.googlesource.com/chromium/src/+/10e9b59ba1acc764a058b8d152e3825ff7db50ba/third_party/blink/renderer/devtools/front_end/extensions/ExtensionAPI.js#577

I suppose that we could repurpose [`singleCallbackArg` in our polyfill](https://github.com/mozilla/webextension-polyfill/blob/34a6ab45595dd2c73130f2d9893452f6aa622d85/src/browser-polyfill.js#L95), so that:

- true = force single value (=current implementation)
- unset = array or single value depending on number of arguments (=current implementation)
- false = force array (=new behavior, to fix this bug).

keithclark:
[...]

PR #170, opened 2019-01-17
test: Use Ajv to test that `api‑metadata.json` matches schema

ExE-Boss:
This adds a&nbsp;JSON&nbsp;Schema for&nbsp;<code>api&#x2011;metadata.json</code> and uses `ajv` and <code>better&#x2011;ajv&#x2011;errors</code> to&nbsp;validate <code>api&#x2011;metadata.json</code> against <code>api&#x2011;metadata.schema.json</code>.

review?(@Rob--W)


api-metadata.schema.json
```diff
@@ -0,0 +1,59 @@
...
+        },
+        "maxArgs": {
+          "description": "The maximum number of arguments which may be passed to the function. If called with more than this number of arguments, the wrapper will raise an exception.",
+          "type": "number"
+        },
+        "fallbackToNoCallback": {
+          "description": "If the function doesn't take a callback parameter in some browsers and throws an error when a callback is passed.",
+          "type": "boolean"
+        },
+        "noCallback": {
```
Rob--W: Why is this here? Internal helper properties in the polyfill itself should not leak to the schema definition of the JSON file.

api-metadata.schema.json
```diff
@@ -0,0 +1,59 @@
...
+          "description": "Set when `fallbackToNoCallback` is set and this browser doesn't support callbacks for this function.\nMUST NOT BE SET IN api-metadata.json.",
+          "type": "boolean"
+        },
+        "singleCallbackArg": {
```
Rob--W: The meaning of this changes in #175 , as documented here : https://github.com/mozilla/webextension-polyfill/issues/168#issuecomment-453623762

Rob--W (changes requested):
Could you also paste examples of the output when schema validation fails?

That could even be encoded in the test, by knowingly passing an invalid API scheme to the validator, and asserting that the reported error matches with the expected error. In this way, we will know for sure that the schema validator is actually strictly enforcing the schema.

api-metadata.schema.json
Rob--W: If you remove the definition here, then validation of the JSON file with API metadata will (should) fail.

ExE-Boss: I currently achieve that with: https://github.com/mozilla/webextension-polyfill/blob/34bdcfc63f63e03f55bf4f15181edafb19f55376/api-metadata.schema.json#L32-L36

But removing it should work too.

Rob--W (commented):
I wrote a comment but forgot to post it. Here it was:

PR #171, opened 2019-01-17
docs: Improve JSDoc of `browser‑polyfill.js` functions and types

ExE-Boss:
A&nbsp;lot&nbsp;of the&nbsp;current JSDoc is&nbsp;straight up&nbsp;wrong or&nbsp;incomplete.

This corrects it&nbsp;to&nbsp;improve intellisense in&nbsp;**VisualStudio&nbsp;Code** when developing this&nbsp;polyfill.

review?(@Rob--W)


src/browser-polyfill.js
```diff
@@ -30,17 +37,26 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
...
-     *        function receives, as its only argument, the key being created.
+     * @template {object} K
+     * @template V
      */
     class DefaultWeakMap extends WeakMap {
+      /**
+       * @param {function(K):V} createItem
+       *        A function which will be called in order to create the value for any
+       *        key which does not exist, the first time it is accessed. The
+       *        function receives, as its only argument, the key being created.
```
ExE-Boss: The `createItem` function documentation was moved to&nbsp;the&nbsp;correct location.

src/browser-polyfill.js
```diff
@@ -30,17 +37,26 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
...
+       */
       constructor(createItem, items = undefined) {
         super(items);
         this.createItem = createItem;
       }
 
+      /**
+       * @param {K} key
+       * @returns {V}
+       */
```
ExE-Boss: This is&nbsp;necessary for&nbsp;proper intellisense.

src/browser-polyfill.js
```diff
@@ -109,19 +124,16 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
      *        The name of the method which is being wrapped.
      * @param {object} metadata
      *        Metadata about the method being wrapped.
-     * @param {integer} metadata.minArgs
+     * @param {number} metadata.minArgs
```
ExE-Boss: This has to&nbsp;be&nbsp;set to&nbsp;`number`, as&nbsp;`integer` doesn’t exist.

src/browser-polyfill.js
```diff
@@ -109,19 +124,16 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
...
      *        The minimum number of arguments which must be passed to the
      *        function. If called with fewer than this number of arguments, the
      *        wrapper will raise an exception.
-     * @param {integer} metadata.maxArgs
+     * @param {number} metadata.maxArgs
      *        The maximum number of arguments which may be passed to the
      *        function. If called with more than this number of arguments, the
      *        wrapper will raise an exception.
-     * @param {integer} metadata.maxResolvedArgs
```
ExE-Boss: `maxResolvedArgs` hasn’t existed for&nbsp;a&nbsp;long time.

src/browser-polyfill.js
```diff
@@ -191,33 +205,46 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
...
+     * @param {object} [metadata]
      *        An object tree containing metadata used to automatically generate
      *        Promise-based wrapper functions for asynchronous. Any function in
      *        the `target` object tree which has a corresponding metadata object
      *        in the same location in the `metadata` tree is replaced with an
      *        automatically-generated wrapper function, as described in
-     *        {@see wrapAsyncFunction}
+     *        {@link wrapAsyncFunction}
      *
-     * @returns {Proxy<object>}
```
ExE-Boss: The result of&nbsp;the&nbsp;`Proxy` constructor doesn’t work this way.

src/browser-polyfill.js
```diff
@@ -297,16 +324,18 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
...
+       * original value of the target if that value is declared read-only and
+       * non-configurable. For this reason, we create an object with the
+       * prototype set to `target` instead of using `target` directly.
+       * Otherwise we cannot return a custom object for APIs that
+       * are declared read-only and non-configurable, such as `chrome.devtools`.
+       *
+       * The proxy handlers themselves will still use the original `target`
+       * instead of the `proxyTarget`, so that the methods and properties are
+       * dereferenced via the original targets.
+       */
```
ExE-Boss: Converted comment to&nbsp;JSDoc.

src/browser-polyfill.js
```diff
@@ -397,10 +430,14 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
...
-        // to translate the message into a resolved promise or a rejected
-        // promise).
+        /**
+         * A small helper to send the message if the promise resolves
+         * and an error if the promise rejects (a wrapped sendMessage has
+         * to translate the message into a resolved promise or a rejected
+         * promise).
+         *
+         * @param {Promise<any>} promise
+         */
```
ExE-Boss: Converted comment to&nbsp;JSDoc.

src/browser-polyfill.js
```diff
@@ -72,20 +88,19 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
...
      *        An object containing the resolution and rejection functions of a
      *        promise.
-     * @param {function} promise.resolve
+     * @param {function((T|T[])):void} promise.resolve
      *        The promise's resolution function.
-     * @param {function} promise.rejection
+     * @param {function(Error):void} promise.reject
      *        The promise's rejection function.
-     * @param {object} metadata
+     * @param {FunctionMetadata} metadata
```
Rob--W: `object`.

src/browser-polyfill.js
```diff
@@ -72,20 +88,19 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
      * - Otherwise, the promise is resolved to an array containing all of the
      *   function's arguments.
      *
+     * @template T
```
Rob--W: Remove `@template T` and its uses below.

src/browser-polyfill.js
```diff
@@ -109,19 +124,16 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
...
+     * @param {number} metadata.maxArgs
      *        The maximum number of arguments which may be passed to the
      *        function. If called with more than this number of arguments, the
      *        wrapper will raise an exception.
-     * @param {integer} metadata.maxResolvedArgs
-     *        The maximum number of arguments which may be passed to the
-     *        callback created by the wrapped async function.
      *
-     * @returns {function(object, ...*)}
+     * @returns {function(object, ...*):Promise}
```
Rob--W: Undo all changes starting from here, until the end of the file. They do not improve the readability of the code.

src/browser-polyfill.js
```diff
@@ -72,20 +88,19 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
...
      *        The promise's rejection function.
-     * @param {object} metadata
+     * @param {FunctionMetadata} metadata
      *        Metadata about the wrapped method which has created the callback.
-     * @param {integer} metadata.maxResolvedArgs
-     *        The maximum number of arguments which may be passed to the
-     *        callback created by the wrapped async function.
      *
-     * @returns {function}
+     * @returns {function(...T):void}
```
Rob--W: undo

src/browser-polyfill.js
Rob--W: Could you elaborate? What's the difference before and after adding this?

The "get" method for WeakMaps has a well-defined meaning, and the lack of annotations will probably not hurt the comprehension of this method, so I prefer the removal of these boilerplate comments.

src/browser-polyfill.js
```diff
@@ -30,17 +37,26 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
...
+     * @template {object} K
+     * @template V
      */
     class DefaultWeakMap extends WeakMap {
+      /**
+       * @param {function(K):V} createItem
+       *        A function which will be called in order to create the value for any
+       *        key which does not exist, the first time it is accessed. The
+       *        function receives, as its only argument, the key being created.
+       * @param {ReadonlyArray<[K, V]>} [items]
```
Rob--W: Remove this `@param`. It doesn't help (perhaps we should remove the param altogether since it is never used - please do NOT make that change in this commit though).

src/browser-polyfill.js
Rob--W: Why? This comment is just explaining the rationale for internal details and not relevant to API consumers.

Rob--W (changes requested):
This PR mixes some useful changes with changes of questionable value. Please make the following changes, and if you strongly disagree, show why the change is necessary anyway. Keeping unnecessary changes minimal makes it easier to use git blame to follow the history of the code.

Some of the changes only change the format of the JSDoc, but since there is no automatic validation of the formatting in the repo, they may go out of date in the future anyway since none of the main developers of this project are using your tools.

src/browser-polyfill.js
```diff
@@ -30,17 +37,26 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
      * not exist when accessed, but behaves exactly as an ordinary WeakMap
      * otherwise.
      *
-     * @param {function} createItem
-     *        A function which will be called in order to create the value for any
-     *        key which does not exist, the first time it is accessed. The
-     *        function receives, as its only argument, the key being created.
+     * @template {object} K
+     * @template V
```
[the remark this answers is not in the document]
ExE-Boss: This is necessary for proper IntelliSense support.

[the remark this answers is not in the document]
Rob--W: What is "proper IntelliSense support"? What happens without these comments? We don't use IntelliSense, so such comments are not likely to be maintained in the long run. Therefore we can as well not add them in the first place.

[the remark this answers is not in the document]
ExE-Boss: Without `@template`, **VS&nbsp;Code** assumes that the&nbsp;`DefaultWeakMap` is&nbsp;untyped and that the&nbsp;result of&nbsp;`get(…)` is&nbsp;`any`, since it&nbsp;can’t determine the&nbsp;values of&nbsp;`K`&nbsp;and&nbsp;`V` of&nbsp;the&nbsp;extended [`WeakMap`](https://github.com/Microsoft/TypeScript/blob/5bcf251f8ae876ae3f4fb12d8b89412d162e23ed/lib/lib.es2015.collection.d.ts#L45-L56(

src/browser-polyfill.js
ExE-Boss: I’ve found a&nbsp;better way to&nbsp;achieve this.

src/browser-polyfill.js
Rob--W: What do you mean by that? Proxies have no prototype, Proxy instances are seemingly transparant wrappers around the original target object. Hence for human readers `Proxy<object>` should be clear enough. Please undo the changes in this whole comment (except maybe for `= {}`).

src/browser-polyfill.js
```diff
@@ -191,33 +196,46 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
...
      *        automatically-generated wrapper function, as described in
-     *        {@see wrapAsyncFunction}
+     *        {@link wrapAsyncFunction}
      *
-     * @returns {Proxy<object>}
+     * @returns {T}
+     *        A Proxy object for the given target.
      */
     const wrapObject = (target, wrappers = {}, metadata = {}) => {
+      /** @type {Partial<T>} */
```
Rob--W: Remove this comment.

src/browser-polyfill.js
```diff
@@ -355,18 +378,19 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
...
+       * @param {function(any):void} sendResponse
        *        A callback which, when called with an arbitrary argument, sends
        *        that value as a response.
+       *
        * @returns {boolean}
        *        True if the wrapped listener returned a Promise, which will later
        *        yield a response. False otherwise.
        */
-      return function onMessage(message, sender, sendResponse) {
+      return function onMessage(message, sender, /** @type {(response?: any) => void} */ sendResponse) {
```
Rob--W: The signature of `sendResponse` is already in the jsdoc; remove this duplicate comment.

src/browser-polyfill.js
```diff
@@ -487,8 +540,8 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
     };
     const settingMetadata = {
       clear: {minArgs: 1, maxArgs: 1},
-      get: {minArgs: 1, maxArgs: 1},
-      set: {minArgs: 1, maxArgs: 1},
+      get:   {minArgs: 1, maxArgs: 1},
+      set:   {minArgs: 1, maxArgs: 1},
```
Rob--W: Undo whitespace changes. We only have one space between `:` and the value.

Rob--W (commented):
Another round of review. Please keep in mind that changes themselves need to be generally useful. We won't accept non-idiomatic changes that only exist to support your IDE. More comments = higher maintenance, and without automatic integration, these comments may eventually go out of date again.

src/browser-polyfill.js
```diff
@@ -170,16 +173,18 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
...
-     * @param {function} method
+     * @param {M} method
      *        The method being wrapped. This is used as the target of the Proxy
      *        object which is created to wrap the method.
      * @param {function} wrapper
      *        The wrapper function which is called in place of a direct invocation
      *        of the wrapped method.
      *
-     * @returns {Proxy<function>}
+     * @returns {M}
```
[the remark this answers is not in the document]
ExE-Boss: Except that `Proxy<function>` is&nbsp;invalid.

src/browser-polyfill.js
ExE-Boss: Except that the JSDoc seems to&nbsp;apply to&nbsp;`return`, not&nbsp;`onMessage`

src/browser-polyfill.js
```diff
@@ -191,33 +196,46 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
...
     };
 
+    /**
+     * Determines whether an object has a property with the specified name.
+     *
+     * @param {any} target The object to test.
+     * @param {string | number | symbol} v A property name.
+     *
+     * @type {(target: any, v: string | number | symbol) => boolean}
+     */
```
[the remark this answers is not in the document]
ExE-Boss: I just copied the definition from TypeScript:
https://github.com/Microsoft/TypeScript/blob/master/lib/lib.es5.d.ts#L137

src/browser-polyfill.js
ExE-Boss: I mean this:
https://github.com/Microsoft/TypeScript/blob/master/lib/lib.es2015.proxy.d.ts#L40

src/browser-polyfill.js
Rob--W: The `@param` should be removed. My comment might have been ambiguous, when I said "please do not make that change", I was referring to the removal of the `items` parameter from the function.

src/browser-polyfill.js
Rob--W: Seems more like a bug in the JSDoc parser. Please remove the inline type definition anyway.

src/browser-polyfill.js
```diff
@@ -440,6 +465,31 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
...
+     *   resolved to that value.
+     * - Otherwise, the promise is resolved to an array containing all of the
+     *   function's arguments.
+     *
+     * @param {object} promise
+     *        An object containing the resolution and rejection functions of a
+     *        promise.
+     * @param {function(any):void} promise.resolve
+     *        The promise's resolution function.
+     * @param {function(any):void} promise.reject
```
Rob--W: I'm aware that these names are likely inspired by the current naming of `makeCallback`, but the name is misleading. Please rename `promise` to `deferred` (since elsewhere we already use the `promise` name for actual promises).

If you prefer, you can leave the current name as it is and make a separate PR to have consistent naming in the whole file.

Rob--W (commented):
The comments are getting in a better shape now. I'm still not too thrilled with the conversion from`Proxy<object>` and `Proxy<function>` to generics T and M. Since your goal is apparently to assist automatic parsing of the JSDoc rules, using TypeScript semantics, could you add a verification step to the build system that verifies that the JSDoc is valid? That's the only way to make sure that the work from this PR is not undone in later changes.

If that is not something that you'd like to do, then I would prioritize human readability over machine-readability, and prefer the use of `Proxy<...>` instead of generics.

src/browser-polyfill.js
```diff
@@ -440,6 +468,31 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
...
+     *        The promise's rejection function.
+     * @param {object} [reply]
```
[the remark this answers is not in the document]
ExE-Boss: That breaks the description of `reply.__mozWebExtensionPolyfillReject__` and `reply.message`.

src/browser-polyfill.js
ExE-Boss: Let’s do that in a&nbsp;latter PR.

ExE-Boss:
JSDoc validation is&nbsp;already performed by&nbsp;eslint: https://github.com/mozilla/webextension-polyfill/blob/949f08cc7d622145a316012aa2490d385072d79d/.eslintrc#L19-L32

src/browser-polyfill.js
ExE-Boss: That violates the&nbsp;<code>valid&#x2011;jsdoc</code> ESLint rule.

ExE-Boss (commented):
JSDoc validation is&nbsp;already performed by&nbsp;ESLint: https://github.com/mozilla/webextension-polyfill/blob/949f08cc7d622145a316012aa2490d385072d79d/.eslintrc#L19-L32

Note that the&nbsp;only way full validity of&nbsp;the&nbsp;JSDoc comments could be&nbsp;enforced would be&nbsp;to&nbsp;migrate the&nbsp;code to&nbsp;TypeScript and&nbsp;then use a&nbsp;direct (esnext&nbsp;to&nbsp;esnext) conversion, but&nbsp;that’s out of&nbsp;scope of&nbsp;this&nbsp;PR.

Rob--W (commented):
If you're not going to add automated verification, how can we verify that the JSDoc is in fact valid (without installing a full-blown IDE...)? The `valid-jsdoc` eslint rule isn't helping, because it accepts the JSDoc before and after your patch. So if in your eyes the JSDoc was invalid before the patch, then there should be some validator that shows a failing validation before, and a passing validation after the patch.

ExE-Boss:
Technically speaking, JSDoc can’t really be&nbsp;validated as&nbsp;it’s just documentation for&nbsp;developers based&nbsp;on&nbsp;[the&nbsp;**Javadoc**&nbsp;syntax](https://en.wikipedia.org/wiki/Javadoc).

src/browser-polyfill.js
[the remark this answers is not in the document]
Rob--W: Why is this using `@type` rather than `@returns {boolean}` ?

Rob--W (commented):
Before this PR, we only used very basic JSDoc annotations (with plain types such as `function` and `string`, with the most complicated type being `Proxy<object>`).
In this PR, you introduced new type annotations based on TypeScript. This raises the bar for contributing to the library, because others are now required to understand TypeScript syntax and its types.

`function` is clearly simpler than `function(any):void`, but the latter is also more precise in the expected types (unsurprisingly). I am willing to accept slightly more complicated syntax if there are clear benefits. Since you've taken the efforts to open this PR and repeatedly respond to reviews, I assume that you see benefits, such as better integration in your IDE. This integration relies on automated tooling, so there must be a way (automatically or manually) for me to verify that your annotations are "correct".

Types such as `ReadonlyArray`, `PropertyKey` and `ProxyHandler<T>` are only meaningful with TypeScript, and I'm not even sure if there is any JSDoc parser that can make sense of `(message, sender: object, sendResponse: (response?: any) => void) => boolean | Promise<any> | void`.

Either show a straightforward way to verify that the types are meaningful, or remove the TypeScript-specific syntax. And if you do decide to drop TypeScript, use `Proxy<object>` and `Proxy<function>` instead of `T` and `M`.

Zearin:
Would love to see this merged.  I like valid JSDoc.  

I don’t like TypeScript, however.  I don’t think it should be included in JSDoc (unlike the source code is written in TypeScript as well).

PR #175, opened 2019-01-31
fix: browser.devtools.inspectedWindow.eval promise should always resolve to an array

### LINKED ISSUE #168: browser.devtools.inspectedWindow.eval() resolves with incorrect promise signature

rpl:
This PR includes the changes needed to ensure that browser.devtools.inspectedWindow.eval returned promise is always resolved to an array (fix #168), and a new integration test to verify it as part of the automated tests.

### DIFF
--- a/src/browser-polyfill.js
+++ b/src/browser-polyfill.js
@@ -93,7 +93,13 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
         if (extensionAPIs.runtime.lastError) {
           promise.reject(extensionAPIs.runtime.lastError);
         } else if (metadata.singleCallbackArg || callbackArgs.length <= 1) {
-          promise.resolve(callbackArgs[0]);
+          if (metadata.singleCallbackArg === false) {
+            // Pass the callback arguments as an array even if its length is <= 1
+            // when explicitly requested by the api metadata.
+            promise.resolve(callbackArgs);
+          } else {
+            promise.resolve(callbackArgs[0]);
+          }
         } else {
           promise.resolve(callbackArgs);
         }
--- a/src/browser-polyfill.js
+++ b/src/browser-polyfill.js
@@ -92,14 +92,9 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
       return (...callbackArgs) => {
         if (extensionAPIs.runtime.lastError) {
           promise.reject(extensionAPIs.runtime.lastError);
-        } else if (metadata.singleCallbackArg || callbackArgs.length <= 1) {
-          if (metadata.singleCallbackArg === false) {
-            // Pass the callback arguments as an array even if its length is <= 1
-            // when explicitly requested by the api metadata.
-            promise.resolve(callbackArgs);
-          } else {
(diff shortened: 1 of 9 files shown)


src/browser-polyfill.js
```diff
@@ -93,7 +93,13 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
         if (extensionAPIs.runtime.lastError) {
           promise.reject(extensionAPIs.runtime.lastError);
         } else if (metadata.singleCallbackArg || callbackArgs.length <= 1) {
-          promise.resolve(callbackArgs[0]);
+          if (metadata.singleCallbackArg === false) {
+            // Pass the callback arguments as an array even if its length is <= 1
+            // when explicitly requested by the api metadata.
```
Rob--W: Collapse clause in parent `if`-condition: `callbackArgs.length <= 1 && metadata.singleCallbackArg !== false`

test/fixtures/devtools-extension/devtools_page.js
```diff
@@ -0,0 +1,14 @@
+console.log("devtools page loaded");
+
+browser.runtime.onMessage.addListener(async msg => {
+  switch (msg.apiMethod) {
+    case "devtools.inspectedWindow.eval": {
+      const evalResult = await browser.devtools.inspectedWindow.eval(...msg.params);
+      return Promise.resolve({evalResult});
```
Rob--W: `return {evalResult};` should be sufficient, as you already have `async` above.

test/fixtures/devtools-extension/devtools_page.js
```diff
@@ -0,0 +1,14 @@
...
+    }
+  }
+
+  return Promise.reject(`devtools_page received an unxpected message: ${msg}`);
```
Rob--W: `throw new Error(...);` because the function is already `async`.

PR #177, opened 2019-02-05
test(devtools): Fix intermittent DevTools API test timeout

ExE-Boss:
This fixes the intermittent timeout issue that caused a #171 test to fail.

## External links:
- https://travis-ci.org/mozilla/webextension-polyfill/builds/487787360#L950

### DIFF
--- a/test/fixtures/devtools-extension/content.js
+++ b/test/fixtures/devtools-extension/content.js
@@ -1,4 +1,6 @@
-test("devtools.inspectedWindow.eval resolved with an error result", async (t) => {
+test("devtools.inspectedWindow.eval resolved with an error result", {
+  timeout: 5000,
+}, async (t) => {
   const {evalResult} = await browser.runtime.sendMessage({
     apiMethod: "devtools.inspectedWindow.eval",
     params: ["throw new Error('fake error');"],
@@ -14,7 +16,9 @@ test("devtools.inspectedWindow.eval resolved with an error result", async (t) =>
        "the second element value property should include the expected error message");
 });
 
-test("devtools.inspectedWindow.eval resolved without an error result", async (t) => {
+test("devtools.inspectedWindow.eval resolved without an error result", {
+  timeout: 5000,
+}, async (t) => {
   const {evalResult} = await browser.runtime.sendMessage({
     apiMethod: "devtools.inspectedWindow.eval",
     params: ["[document.documentElement.localName]"],
--- a/test/fixtures/tape-standalone.js
+++ b/test/fixtures/tape-standalone.js
@@ -9,50 +9,31 @@ if (navigator.userAgent.includes("Chrome/")) {
   browser = "Firefox";
 }
 
-function getArgs(/* desc, options, fn */) {
-  let desc = "(anonymous)";
-  // By setting `timeout` to `0` here, we are allowing V8 to consider
-  // it a to have a signature of `{timeout: number}` instead of `{}`
-  // with a `timeout` property, which is easier on the C++ portion.
-  let options = {timeout: 0};
-  /** @type {function(import("tape").Test):void|Promise<void>} */
-  let fn;
-
-  for (let i = 0, arg; i < arguments.length; i++) {
-    arg = arguments[i];
-    switch (typeof arg) {
-      case "string":
-        desc = arg;
-        break;
-      case "function":
-        fn = arg;
-        break;
-      case "object":
-        Object.assign(options, arg);
-        break;
-    }
-  }
-
-  if (!options.timeout) {
-    options.timeout = DEFAULT_TIMEOUT;
-  }
-
-  return {
-    desc,
-    options,
-    fn,
-  };
-}
-
 // Export as a global a wrapped test function which enforces a timeout by default.
 // eslint-disable-next-line
 /**
- * @param {string} [desc]
- * @param {object} [options]
+ * @param {string} desc
+ * @param {import("tape").TestOptions|null} [options]
  * @param {(test: import("tape").Test) => void|Promise<void>} fn
  */
 window.test = (desc, options, fn) => {
-  ({desc, options, fn} = getArgs(desc, options, fn));
+  if (typeof options === "function") {
+    // Allow swapping options with fn
+    if (typeof fn === "object" && fn) {
+      let tmp = fn;
+      fn = options;
+      options = tmp;
+    } else {
+      fn = options;
+      options = undefined;
+    }
+  }
+
+  options = {
+    timeout: DEFAULT_TIMEOUT,
+    ...options,
+  };
+
   tape(`${desc} (${browser})`, options, async (t) => {
     await fn(t);
   });
--- a/test/fixtures/tape-standalone.js
+++ b/test/fixtures/tape-standalone.js
@@ -15,7 +15,7 @@ function getArgs(/* desc, options, fn */) {
   // it a to have a signature of `{timeout: number}` instead of `{}`
   // with a `timeout` property, which is easier on the C++ portion.
   let options = {timeout: 0};
-  /** @type {function(any):void|Promise<void>} */
+  /** @type {function(import("tape").Test):void|Promise<void>} */
   let fn;
 
   for (let i = 0, arg; i < arguments.length; i++) {
@@ -33,7 +33,7 @@ function getArgs(/* desc, options, fn */) {
     }
   }
 
-  if (!options.timeout || options.timeout < DEFAULT_TIMEOUT) {
+  if (!options.timeout) {
     options.timeout = DEFAULT_TIMEOUT;
   }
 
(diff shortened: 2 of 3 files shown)


Rob--W:
Your patch doesn't work because the `test` function is defined here, without return value: https://github.com/mozilla/webextension-polyfill/blob/949f08cc7d622145a316012aa2490d385072d79d/test/fixtures/tape-standalone.js#L12-L18

To fix this, I think that you can use the `timeout` option instead of explicitly calling `timeoutAfter`, see https://github.com/substack/tape/blob/v4.5.0/readme.markdown#testname-opts-cb and `_timeout` / `timeoutAfter` in https://github.com/substack/tape/blob/v4.5.0/lib/test.js

test/fixtures/devtools-extension/content.js
```diff
@@ -1,4 +1,6 @@
 test("devtools.inspectedWindow.eval resolved with an error result", async (t) => {
+  t.timeoutAfter(5000);
```
Rob--W: This won't work. If you look at the implementation that I linked from my previous comment, then you'll see that `timeoutAfter` is implemented by installing a non-cancelable timer that times out the test. Since our `test` wrapper already calls `timeoutAfter`, the only way to increase the timeout beyond 500ms is to update the `test` wrapper, using the instructions from my comment.

test/fixtures/tape-standalone.js
```diff
@@ -10,8 +10,12 @@ if (navigator.userAgent.includes("Chrome/")) {
...
 // Export as a global a wrapped test function which enforces a timeout by default.
-window.test = (desc, fn) => {
-  tape(`${desc} (${browser})`, async (t) => {
+window.test = (desc, options, fn) => {
+  if (!fn && typeof options === "function") {
+    fn = options;
+    options = undefined;
+  }
+  tape(`${desc} (${browser})`, options, async (t) => {
     t.timeoutAfter(DEFAULT_TIMEOUT);
```
Rob--W: You still need to remove the `timeoutAfter` call. Otherwise you cannot have a shorter timeout (there would be two timeouts).

So, declare an `options` object unconditionally with default value `{timeout: DEFAULT_TIMEOUT}`, and use `Object.assign` to merge the input `options` parameter.

And to avoid reshuffling of arguments, let's put `options` at the third place, after `fn`. Although `tape` doesn't document it, the parameters can be in any order: https://github.com/substack/tape/blob/v4.5.0/lib/test.js#L18-L37

Rob--W:
@ExE-Boss Have you already made some changes? The PR hasn't been updated yet.

test/fixtures/tape-standalone.js
ExE-Boss: I’ve now implemented this using a&nbsp;function similar to&nbsp;the&nbsp;one used by&nbsp;`tape`.

test/fixtures/tape-standalone.js
```diff
@@ -9,10 +9,50 @@ if (navigator.userAgent.includes("Chrome/")) {
...
+        fn = arg;
+        break;
+      case "object":
+        Object.assign(options, arg);
+        break;
+    }
+  }
+
+  if (!options.timeout || options.timeout < DEFAULT_TIMEOUT) {
+    options.timeout = DEFAULT_TIMEOUT;
```
Rob--W: This implementation means that a test cannot set a lower timeout. That seems like an unnecessary restriction.

test/fixtures/tape-standalone.js
```diff
@@ -9,10 +9,50 @@ if (navigator.userAgent.includes("Chrome/")) {
...
-window.test = (desc, fn) => {
-  tape(`${desc} (${browser})`, async (t) => {
-    t.timeoutAfter(DEFAULT_TIMEOUT);
+/**
+ * @param {string} [desc]
+ * @param {object} [options]
+ * @param {function(any):void|Promise<void>} fn
+ */
+window.test = (desc, options, fn) => {
+  ({desc, options, fn} = getArgs(desc, options, fn));
```
Rob--W: The implementation here is unnecessarily complicated. All you need is to take change the method signature to this:

```javascript
window.test = (desc, fn, options = {}) => {
```

and then set the default timeout, like this:

```javascript
if (!options.timeout) {
  options.timeout = DEFAULT_TIMEOUT;
}
```

The above implementation modifies the input parameter. If you want to avoid modifying the input parameter, then you could make a fresh copy as follows (instead of the `if (!options.timeout) {...}` block):

```javascript
options = {
  timeout: DEFAULT_TIMEOUT,
  ...options,
}
```

ExE-Boss: Yeah, but&nbsp;I&nbsp;prefer putting the&nbsp;options on&nbsp;the&nbsp;second place in&nbsp;the&nbsp;method signature and allow omitting the&nbsp;description, like what tape&nbsp;does.

Rob--W: We aren't using that format, and the resulting code becomes considerably more complicated with it.

Supporting only one format is easier to maintain (YAGNI), so let's stick to that.

test/fixtures/tape-standalone.js
```diff
@@ -9,10 +9,51 @@ if (navigator.userAgent.includes("Chrome/")) {
...
+
 // Export as a global a wrapped test function which enforces a timeout by default.
-window.test = (desc, fn) => {
-  tape(`${desc} (${browser})`, async (t) => {
-    t.timeoutAfter(DEFAULT_TIMEOUT);
+// eslint-disable-next-line
+/**
+ * @param {string} [desc]
+ * @param {object} [options]
+ * @param {(test: import("tape").Test) => void|Promise<void>} fn
```
Rob--W: Let's remove this JSDoc. The names are self-explanatory, and "documentation" is available in the form of existing examples (tests). Anyone who cares can look in this file and see that this is simply a wrapper around `tape`.

ExE-Boss: This&nbsp;is there for&nbsp;intellisense in&nbsp;IDEs.

test/fixtures/tape-standalone.js
ExE-Boss: I prefer putting the options before the test case, as it makes it easier to find them, rather than having to find the end of the test case function.

It also makes it obvious which test cases are using custom options and which ones are not.

test/fixtures/tape-standalone.js
```diff
@@ -10,9 +10,31 @@ if (navigator.userAgent.includes("Chrome/")) {
 }
 
 // Export as a global a wrapped test function which enforces a timeout by default.
-window.test = (desc, fn) => {
-  tape(`${desc} (${browser})`, async (t) => {
-    t.timeoutAfter(DEFAULT_TIMEOUT);
+// eslint-disable-next-line
```
Rob--W: Don't use a blanket eslint disabling. Always be explicit with the linting rule that you intend to turn off. I think that you wanted to disable `valid-jsdoc` here; that exception can be removed when you follow my suggestion of simplifying the parameter type.

test/integration/setup.js
```diff
@@ -192,8 +193,8 @@ const defineExtensionTests = ({description, extensions, skip, openDevTools}) =>
           path.join(__dirname, "..", "fixtures", extensionDirName));
         const srcPolyfill = path.join(__dirname, "..", "..", "dist", "browser-polyfill.js");
 
-        const tmpDir = tmp.dirSync({unsafeCleanup: true});
-        const extensionPath = path.join(tmpDir.name, extensionDirName);
+        tempDir = tmp.dirSync({unsafeCleanup: true});
+        const extensionPath = path.join(tempDir.name, extensionDirName);
```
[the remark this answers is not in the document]
ExE-Boss: This was detected by my IDE.

test/fixtures/tape-standalone.js
```diff
@@ -10,19 +10,23 @@ if (navigator.userAgent.includes("Chrome/")) {
...
   if (typeof options === "function") {
     // Allow swapping options with fn
     if (typeof fn === "object" && fn) {
-      let tmp = fn;
-      fn = options;
-      options = tmp;
+      [fn, options] = [options, fn];
     } else {
       fn = options;
       options = undefined;
```
Rob--W: This branch can be removed too; the logic would be covered by the `[fn, options] = [options, fn];` assignment.

test/fixtures/tape-standalone.js
```diff
@@ -10,19 +10,23 @@ if (navigator.userAgent.includes("Chrome/")) {
...
- * @param {import("tape").TestOptions|null} [options]
- * @param {(test: import("tape").Test) => void|Promise<void>} fn
+ *        The test description
+ * @param {object} [options]
+ *        The test options, can be omitted.
+ * @param {number} [options.timeout=DEFAULT_TIMEOUT]
+ *        The time after which the test fails automatically, unless it has already passed.
+ * @param {boolean} [options.skip]
+ *        Whether the test case should be skipped.
+ * @param {function(tape.Test):void|Promise<void>} fn
```
Rob--W: Remove the return type. `void|Promise<void>` is ambiguous (as the one can also read it as `fn = function or Promise`). To cover the return type, you can change the description to  `The (optionally async) function that runs the test case`.

ExE-Boss: I ended up putting the&nbsp;return type inside a&nbsp;parenthesis to&nbsp;make it&nbsp;clearer (even though it&nbsp;doesn’t matter).

PR #188, opened 2019-05-29
chore: Add `.editorconfig`

ExE-Boss:
This adds an [EditorConfig] file, so that the project’s indentation style is maintained without needing to reconfigure the editor.

---

[EditorConfig]: https://editorconfig.org

review?(@Rob--W)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+root = true
+
+[*]
+end_of_line = lf
+charset = utf-8
+indent_style = space
+indent_size = 2
+insert_final_newline = true
+trim_trailing_whitespace = true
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,9 @@
+root = true
+
+[*]
```
Rob--W: `*` -> `*.{html,js,json}`

ExE-Boss: So that `utf‑8` encoding and `LF` line endings are enforced everywhere.

ExE-Boss:
@Rob--W
The Travis build passed, but the commit status hasn’t been correctly updated.

Issue #116, opened 2018-05-21
Polyfill the runtime.getBrowserInfo() function on browsers which don’t support it


ExE-Boss:
The [`runtime.getBrowserInfo()` function](https://developer.mozilla.org/Add-ons/WebExtensions/API/runtime/getBrowserInfo) is [currently only supported in Firefox](https://developer.mozilla.org/Add-ons/WebExtensions/API/runtime/getBrowserInfo#Browser_compatibility).

Partial support for it can be added to other browsers by using user agent sniffing (I know, sniffing has horrible reputation) to extract the version and browser information.

User agent sniffing can be done in JavaScript using the [`useragent` package <img src="https://img.shields.io/npm/v/useragent.svg" alt="(on npm)" align="top"/>](https://www.npmjs.com/package/useragent) ([Test With RunKit](https://runkit.com/embed/apx79wofdew2)).

ExE-Boss:
@lidel Fixed it 👍🏻.

Rob--W:
The polyfill only offers access to APIs (via a nice Promise-based API) that exist in the browser. It does not provide polyfills for APIs that are missing from the platform: https://github.com/mozilla/webextension-polyfill#api-methods-or-options-that-are-only-available-when-running-in-firefox

PR #193, opened 2019-07-23
Suggest polyfill for contentScripts.register

fregante:
While polyfilling the actual APIs in Chrome is out of scope, it might make sense suggesting that there are other packages available to fill the gap.

### DIFF
--- a/README.md
+++ b/README.md
@@ -299,7 +299,9 @@ When an extension that uses this library doesn't behave as expected on Firefox, 
 
 ### API methods or options that are only available when running in Firefox
 
-This library does not provide any polyfill for API methods and options that are only available on Firefox, and they are actually considered out of the scope of this library.
+This library does not provide any polyfill for API methods and options that are only available on Firefox, and they are actually considered out of the scope of this library. There may however be polyfills available for Chrome, such as:
+
+- [`contentScripts.register()` polyfill](https://github.com/bfred-it/content-scripts-register-polyfill)
 
 ### tabs.executeScript
 


Rob--W:
Polyfilling missing functionality is outside the scope of this project, so we will not be recommending alternatives from the README. Some functionality cannot be polyfilled, and `contentScripts.register` is an example of that. There are several edge cases where a `tabs.executeScript`-based polyfill behaves differently from `contentScripts.register` (e.g. with `runAt: "document_start"`).

fregante:
> There are several edge cases where a `tabs.executeScript`-based polyfill behaves differently from `contentScripts.register` (e.g. with `runAt: "document_start"`).

Right, but that'd be a note to put on my repo. I hope you can agree that using a polyfill that works in most cases is better than having people having to hack their way around `executeScript`, which has the same limitations and still doesn't say that (e.g. `runAt: "document_start"` can't guarantee that)

Rob--W:
Chrome lacks the primitives to build a proper polyfill for `contentScripts.register`, so users of `contentScripts.register` should look into the alternatives that are the best fit for them. Whether that's a polyfill with known limitations, a different API, or a completely different extension implementation is up to them. Chrome's `declarativeContent.RequestContentScript` API has many defects, but in some specific cases (`runAt` not important, static files only, JS only), may be the best alternative.

As said, polyfills for functionality are [outside the scope of this library](https://github.com/mozilla/webextension-polyfill/blob/51998c1c64ca09aba30efa544c0447a42508d31b/README.md#api-methods-or-options-that-are-only-available-when-running-in-firefox). There are APIs that are much easier to polyfill than `contentScripts.register`, and they have also been closed (e.g. #116). Nevertheless, since you appear to be interested in trying to perfect the polyfill, here is some feedback:

- Your library requires the `"tabs"` permission, host permissions and `optional_permissions`. This is not documented.
- The library is susceptible to running scripts in the wrong document, if the tab navigates during the execution of the `tabs.onUpdated` function.
- The library's match pattern parser does not behave as specified (e.g. there is no such thing as single-character wildcards, `/?.+` should not have a `?`; `'.+'` in path wildcards should be `'.*`, slashes do not need to be escaped, `<all_urls>` is not accounted for)
- The library is unable to run scripts in specific frames only, as script injection is based on the top-level URL.
- The library may cause log spam when a URL matches a pattern, but the browser still refuses to execute scripts (e.g. in the Chrome Web Store domain).

As you might see, providing a *generic* polyfill for `contentScripts.register` is difficult, and writing code for specific scenarios is usually more performant than using a polyfill.

fregante:
> There are APIs that are much easier to polyfill than `contentScripts.register`, and they have also been closed (e.g. #116

To be clear, I'm not suggesting to include the polyfill in `webextension-polyfill` though; only to add some suggestions in the readme.

> here is some feedback:

Thank you! <3 I'll look into those issues.

> Your library requires the `"tabs"` permission, host permissions and `optional_permissions`. This is not documented.

It doesn't require `tabs` and `optional_permissions` for sure. My understanding is that `contentScript.register` needs permissions to the hosts you're trying to register the scripts to. How else would it work in Firefox? You can't register scripts to hosts you can't access.

I haven't tested what happens when the user tries to register a script on a non-allowed domain. I'll open an issue about that specifically, it should match the API's behavior.

fregante:
> writing code for specific scenarios is usually more performant than using a polyfill.

Yet we are on `mozilla/webextension-polyfill` while it's more performant to use the unwrapped `chrome.*` APIs 😜

Rob--W:
> > Your library requires the `"tabs"` permission, host permissions and `optional_permissions`. This is not documented.
> 
> It doesn't require `tabs` and `optional_permissions` for sure. My understanding is that `contentScript.register` needs permissions to the hosts you're trying to register the scripts to.

The "url" property of a tab is only available with the `"tabs"` permission. You're reading  `.url` at https://github.com/fregante/content-scripts-register-polyfill/blob/2202738946b6da5eddcab672340e97f09313b13b/index.ts#L69

The `chrome.permissions` API is only available if the `optional_permissions` field is set in manifest.json. You're using `chrome.permissions.contains` at https://github.com/fregante/content-scripts-register-polyfill/blob/2202738946b6da5eddcab672340e97f09313b13b/index.ts#L33

> I haven't tested what happens when the user tries to register a script on a non-allowed domain. I'll open an issue about that specifically, it should match the API's behavior.

If your code reaches `wasPreviouslyLoaded`, but is not allowed to execute (e.g. the Chrome Web Store), then an error will be returned by `chrome.tabs.executeScript`, and propagate up to https://github.com/fregante/content-scripts-register-polyfill/blob/2202738946b6da5eddcab672340e97f09313b13b/index.ts#L75


> > writing code for specific scenarios is usually more performant than using a polyfill.
> 
> Yet we are on `mozilla/webextension-polyfill` while it's more performant to use the unwrapped `chrome.*` APIs stuck_out_tongue_winking_eye

The cost of a promise wrapper is minimal. The cost of setting up listeners, and repeatedly communicating between the browser process, extension process and content processes is non-negligible.

fregante:
> The "url" property of a tab is only available with the `"tabs"` permission. You're reading `.url` at [fregante/content-scripts-register-polyfill:index.ts@`2202738`#L69](https://github.com/fregante/content-scripts-register-polyfill/blob/2202738946b6da5eddcab672340e97f09313b13b/index.ts#L69)

In Firefox yes, but not in Chrome. This is a polyfill for Chrome. I tested it there and `url` appears when you have permissions to the host; I think it's even there with a simple `activeTab` permission when it's activated.

PR #196, opened 2019-07-29
feat: add type declaration file for TypeScript

t7yang:
Add `index.d.ts` for typing is really helpful for TypeScript Developers.
Build and locally install for type checking was tested.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,8 @@
   "description": "A lightweight polyfill library for Promise-based WebExtension APIs in Chrome.",
   "main": "dist/browser-polyfill.js",
   "files": [
-    "dist"
+    "dist",
+    "index.d.ts"
   ],
   "repository": {
     "type": "git",
@@ -49,6 +50,9 @@
     "tape": "^4.9.1",
     "tape-async": "^2.3.0",
     "tmp": "0.0.33"
+  },
+  "peerDependencies": {
+    "@types/firefox-webext-browser": ">=58.0.0"
   },
   "nyc": {
     "reporter": [
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,1 @@
+export default browser;


Rob--W:
If this PR wasn't merged, what alternatives do you have to have a typed "browser" API in your project? What's the purpose of `index.d.ts`?

package.json
```diff
@@ -50,6 +51,9 @@
     "tape-async": "^2.3.0",
     "tmp": "0.0.33"
   },
+  "peerDependencies": {
+    "@types/firefox-webext-browser": ">=58.0.0"
```
fregante: I don't think this is a good idea. Most people don't need these types and `peerDependencies` won't be installed automatically, resulting in a warning.

Either it's a dependency or it's not.

fregante:
> If this PR wasn't merged, what alternatives do you have to have a typed "browser" API in your project? 

```js
npm i @types/firefox-webext-browser -D
```

This works without this PR and is required even with this PR.

t7yang:
> 
> 
> > If this PR wasn't merged, what alternatives do you have to have a typed "browser" API in your project?
> 
> ```js
> npm i @types/firefox-webext-browser -D
> ```
> 
> This works without this PR and is required even with this PR.

What if developer want to import the `browser` from `webextension-polyfill` and don't want to attach `browser` to `window` in Chromium based runtime. Not everyone want to attach something to `globalThis`.

Besides, it just a types declaration file, didn't cost to much.

package.json
t7yang: For those don't need it, we should not force them to install and they can ignore the warning if their project work without types. For those who need it, we should give them a sign (Hey look like you missing something you might need to install).

fregante:
> What if developer want to import the `browser` from `webextension-polyfill` and don't want to attach `browser` to `window` in Chromium based runtime

This PR does not prevent that. It's specified as a `peerDependency` so it has _no_ effect on how TypeScript will load it.

Change it to a `dependency` and _maybe_ that changes, but I still think it doesn't.

rpl:
We are glad that there types declarations for the WebExtensions API and/or the polyfill, thanks a lot for that, nevertheless @Rob--W and I agree with @fregante, we don't see this as something that should be part of the polyfill dependencies.

PR #200, opened 2019-08-01
Removed runtime.getBrowserInfo() 

### LINKED ISSUE #184: runtime.getBrowserInfo() should not be in api-metadata.json
This method is not supported on Chrome.

Apparently it was already supposed to be removed in https://github.com/mozilla/webextension-polyfill/pull/125?

Shubhamchinda:
Fixes #184 

@Rob--W , Let me know if other changes are required. :)

Thanks a lot!

### DIFF
--- a/api-metadata.json
+++ b/api-metadata.json
@@ -203,6 +203,22 @@
       "maxArgs": 1
     }
   },
+  "devtools": {
+    "inspectedWindow": {
+      "eval": {
+        "minArgs": 1,
+        "maxArgs": 2,
+        "singleCallbackArg": false
+      }
+    },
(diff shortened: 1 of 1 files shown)


api-metadata.json
```diff
@@ -422,10 +404,6 @@
       "minArgs": 0,
       "maxArgs": 0
     },
-    "getBrowserInfo": {
```
Rob--W: This is the only change that should be kept. All other removals from this PR should be undone.

Shubhamchinda: Ohh Sorry! 
I've updated it.

Issue #204, opened 2019-10-14
Half of the chrome privacy API is not available


Manvel:
Just noticed that:
```
browser.privacy.websites.thirdPartyCookiesAllowed.get({});
```

Throws the error below:
```
Uncaught TypeError: Error in invocation of types.ChromeSetting.get(object details, function callback): No matching signature.
```

I'm new to current polyfill library, but having a rough look into the [src/browser-polyfill.js](https://github.com/mozilla/webextension-polyfill/blob/master/src/browser-polyfill.js#L494) seems like there are some privacy APIs that [are available in Chrome](https://developer.chrome.com/extensions/privacy) which are not supported by `webextension-polyfill`.

Not sure if that is intentional that not all Chrome APIs are supported, but if Chrome API support can be ensured, that will help me a lot making my Chrome Extension also being available in Firefox.

Rob--W:
Support for the privacy namespace in the polyfill was introduced in #133, and APIs that weren't supported in Firefox were intentionally left out, because extensions that use those APIs are not cross-browser compatible.

So far, the scope of the polyfill has been to only provide a polyfill to APIs that are implemented by Firefox. I'm not opposed to supporting promises in the full `privacy` namespace if there are good reasons for it. If we really want to support promises in any part of the API namespace, then I wonder whether we should have a generic wrapper for any property in the `privacy.{network,services,websites}` namespace without explicitly enumerating all properties.

> Not sure if that is intentional that not all Chrome APIs are supported, but if Chrome API support can be ensured, that will help me a lot making my Chrome Extension also being available in Firefox.

How would providing a promisified version of the API help you with making the extension available to Firefox, if those APIs are only supported in Chrome?

Manvel:
> Support for the privacy namespace in the polyfill was introduced in #133, and APIs that weren't supported in Firefox were intentionally left out, because extensions that use those APIs are not cross-browser compatible.

I see, hmm waiting Chrome switching into using promises don't seem to be something we can expect unfortunately anytime soon :( Same seem to be true for using `Browser.*` instead of `Chrome.*`.

> So far, the scope of the polyfill has been to only provide a polyfill to APIs that are implemented by Firefox. I'm not opposed to supporting promises in the full privacy namespace if there are good reasons for it.

[...]

PR #205, opened 2019-10-14
fix: Added missing chrome privacy api settings

Manvel:
This PR is for fixing https://github.com/mozilla/webextension-polyfill/issues/204

### DIFF
--- a/src/browser-polyfill.js
+++ b/src/browser-polyfill.js
@@ -492,29 +492,19 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
       set: {minArgs: 1, maxArgs: 1},
     };
 
-    // Wrap all "network", "services", "websites" APIs in chrome.privacy.*
-    const privacyTypes = ["network", "services", "websites"];
-    const privacy = {
-      get supported() {
-        return extensionAPIs.privacy;
-      },
-    };
-    for (const privacyType of privacyTypes) {
-      Object.defineProperty(privacy, privacyType, {get: () => {
-        if (extensionAPIs.privacy[privacyType]) {
-          return extensionAPIs.privacy[privacyType];
+    apiMetadata.privacy = {};
+    for (const type of ["network", "services", "websites"]) {
+      Object.defineProperty(apiMetadata.privacy, type, {get: () => {
+        const origPrivacyType = extensionAPIs.privacy && extensionAPIs.privacy[type];
+        if (origPrivacyType) {
+          const privacyType = {};
+          for (const privacyName of Object.keys(origPrivacyType)) {
+            privacyType[privacyName] = settingMetadata;
+          }
+          apiMetadata.privacy[type] = privacyType;
+          return privacyType;
         }
-        return {};
       }});
-    }
-    if (privacy.supported) {
-      apiMetadata.privacy = {};
-      for (const privacyType of privacyTypes) {
-        apiMetadata.privacy[privacyType] = {};
-        for (const privacyName of Object.keys(privacy[privacyType])) {
-          apiMetadata.privacy[privacyType][privacyName] = settingMetadata;
-        }
-      }
     }
 
     return wrapObject(extensionAPIs, staticWrappers, apiMetadata);
--- a/src/browser-polyfill.js
+++ b/src/browser-polyfill.js
@@ -507,6 +507,7 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
           }
         },
         set: (value) => value,
+        configurable: true,
       });
     }
 
--- a/src/browser-polyfill.js
+++ b/src/browser-polyfill.js
@@ -491,33 +491,14 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
       get: {minArgs: 1, maxArgs: 1},
       set: {minArgs: 1, maxArgs: 1},
     };
-    apiMetadata.privacy = {
-      network: {
-        networkPredictionEnabled: settingMetadata,
-        webRTCMultipleRoutesEnabled: settingMetadata,
-        webRTCNonProxiedUdpEnabled: settingMetadata,
-        webRTCIPHandlingPolicy: settingMetadata,
-      },
-      services: {
-        alternateErrorPagesEnabled: settingMetadata,
-        autofillEnabled: settingMetadata,
-        autofillAddressEnabled: settingMetadata,
-        autofillCreditCardEnabled: settingMetadata,
-        passwordSavingEnabled: settingMetadata,
-        safeBrowsingEnabled: settingMetadata,
-        safeBrowsingExtendedReportingEnabled: settingMetadata,
-        searchSuggestEnabled: settingMetadata,
-        spellingServiceEnabled: settingMetadata,
-        translationServiceEnabled: settingMetadata,
-      },
-      websites: {
-        thirdPartyCookiesAllowed: settingMetadata,
-        hyperlinkAuditingEnabled: settingMetadata,
-        referrersEnabled: settingMetadata,
-        doNotTrackEnabled: settingMetadata,
-        protectedContentEnabled: settingMetadata,
-      },
-    };
+    // Wrap all "network", "services", "websites" APIs in chrome.privacy.*
+    apiMetadata.privacy = {};
+    for (const privacyType of ["network", "services", "websites"]) {
+      apiMetadata.privacy[privacyType] = {};
+      for (const privacyName in extensionAPIs.privacy[privacyType]) {
+        apiMetadata.privacy[privacyType][privacyName] = settingMetadata;
+      }
+    }
 
     return wrapObject(extensionAPIs, staticWrappers, apiMetadata);
   };
--- a/src/browser-polyfill.js
+++ b/src/browser-polyfill.js
@@ -494,17 +494,20 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
 
     apiMetadata.privacy = {};
     for (const type of ["network", "services", "websites"]) {
-      Object.defineProperty(apiMetadata.privacy, type, {get: () => {
-        const origPrivacyType = extensionAPIs.privacy && extensionAPIs.privacy[type];
-        if (origPrivacyType) {
-          const privacyType = {};
-          for (const privacyName of Object.keys(origPrivacyType)) {
-            privacyType[privacyName] = settingMetadata;
+      Object.defineProperty(apiMetadata.privacy, type, {
+        get: () => {
+          const origPrivacyType = extensionAPIs.privacy && extensionAPIs.privacy[type];
+          if (origPrivacyType) {
+            const privacyType = {};
+            for (const privacyName of Object.keys(origPrivacyType)) {
+              privacyType[privacyName] = settingMetadata;
+            }
+            apiMetadata.privacy[type] = privacyType;
+            return privacyType;
           }
-          apiMetadata.privacy[type] = privacyType;
-          return privacyType;
-        }
-      }});
+        },
+        set: (value) => value,
+      });
     }
 
     return wrapObject(extensionAPIs, staticWrappers, apiMetadata);
--- a/src/browser-polyfill.js
+++ b/src/browser-polyfill.js
@@ -494,14 +494,28 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
     apiMetadata.privacy = {
       network: {
         networkPredictionEnabled: settingMetadata,
+        webRTCMultipleRoutesEnabled: settingMetadata,
+        webRTCNonProxiedUdpEnabled: settingMetadata,
         webRTCIPHandlingPolicy: settingMetadata,
       },
       services: {
+        alternateErrorPagesEnabled: settingMetadata,
+        autofillEnabled: settingMetadata,
+        autofillAddressEnabled: settingMetadata,
+        autofillCreditCardEnabled: settingMetadata,
         passwordSavingEnabled: settingMetadata,
+        safeBrowsingEnabled: settingMetadata,
+        safeBrowsingExtendedReportingEnabled: settingMetadata,
+        searchSuggestEnabled: settingMetadata,
+        spellingServiceEnabled: settingMetadata,
+        translationServiceEnabled: settingMetadata,
       },
       websites: {
+        thirdPartyCookiesAllowed: settingMetadata,
         hyperlinkAuditingEnabled: settingMetadata,
         referrersEnabled: settingMetadata,
+        doNotTrackEnabled: settingMetadata,
+        protectedContentEnabled: settingMetadata,
       },
     };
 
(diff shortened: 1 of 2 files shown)


src/browser-polyfill.js
```diff
@@ -491,19 +491,16 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
...
-        hyperlinkAuditingEnabled: settingMetadata,
-        referrersEnabled: settingMetadata,
-      },
-    };
+    // Wrap all "network", "services", "websites" APIs in chrome.privacy.*
+    if (extensionAPIs.privacy) {
+      apiMetadata.privacy = {};
+      for (const privacyType of ["network", "services", "websites"]) {
+        apiMetadata.privacy[privacyType] = {};
+        for (const privacyName in extensionAPIs.privacy[privacyType]) {
```
Rob--W: This would also include enumerable properties from `Object.prototype`, which shouldn't be exported.

src/browser-polyfill.js
```diff
@@ -491,19 +491,16 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
...
-      services: {
-        passwordSavingEnabled: settingMetadata,
-      },
-      websites: {
-        hyperlinkAuditingEnabled: settingMetadata,
-        referrersEnabled: settingMetadata,
-      },
-    };
+    // Wrap all "network", "services", "websites" APIs in chrome.privacy.*
+    if (extensionAPIs.privacy) {
```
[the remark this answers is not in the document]
Manvel: Good point, will update.

src/browser-polyfill.js
Manvel: Good point, will update.
Guess using `Object.keys()` might be better.

src/browser-polyfill.js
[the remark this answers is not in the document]
Manvel: Updated, I have created separated getters for each of the `privacyTypes` as you have asked, but I feel like that implementation below is simpler:
```js
// Wrap all "network", "services", "websites" APIs in chrome.privacy.*
    const privacy = {
      get supported() {
        return extensionAPIs.privacy;
      },
    };
    if (privacy.supported) {
      apiMetadata.privacy = {};
      for (const privacyType of ["network", "services", "websites"]) {
        apiMetadata.privacy[privacyType] = {};
        for (const privacyName of Object.keys(privacy[privacyType])) {
          apiMetadata.privacy[privacyType][privacyName] = settingMetadata;
        }
      }
    }
```

What you think?

src/browser-polyfill.js
Manvel: Done.

Manvel:
@Rob--W I have addressed the comments, but seems like there is some problems with ChromeDriver `SessionNotCreatedError: session not created: This version of ChromeDriver only supports Chrome version 76`.

Most probably I won't be available tomorrow, but will ensure to address anything needed the day after.

src/browser-polyfill.js
```diff
@@ -491,12 +491,27 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
...
+    };
+    for (const privacyType of privacyTypes) {
+      Object.defineProperty(privacy, privacyType, {get: () => {
+        if (extensionAPIs.privacy[privacyType]) {
+          return extensionAPIs.privacy[privacyType];
+        }
+        return {};
+      }});
+    }
+    if (privacy.supported) {
```
Rob--W: This is not a lazy getter.

As soon as you dereference the `extensionAPIs.privacy` object, the browser has to initialize the extension bindings for `chrome.privacy` object.

A lazy getter is one where you avoid dereferencing the `extensionAPIs.privacy` object until absolutely necessary.

For example, something like this:

```
apiMetadata.privacy = {
  get network() { // <-- This is a getter. You should probably use Object.defineProperty instead.
    let target = extensionAPIs.privacy && extensionAPIs.privacy.network;
    if (!target) {
      let network = {};
      for (let k of Object.keys(target)) {
       network[k] = settingMetadata;
      }
      this.network = network; // <-- This caches the result, hence the "lazy getter" is run only once.
      return network;
    }
  },
};
```

Manvel: Now I feel so stupid, thanks will update.

Manvel: @Rob--W Updated, hope now it looks better.

Manvel:
> > @Rob--W I have addressed the comments, but seems like there is some problems with ChromeDriver `SessionNotCreatedError: session not created: This version of ChromeDriver only supports Chrome version 76`.
> > Most probably I won't be available tomorrow, but will ensure to address anything needed the day after.
> 
> We should probably bump the chromedriver version again... As done in #199. If you open a pull request in the same format, I'd approve and merge it.

Done -> https://github.com/mozilla/webextension-polyfill/pull/208

src/browser-polyfill.js
```diff
@@ -491,19 +491,24 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
...
-      },
-      websites: {
-        hyperlinkAuditingEnabled: settingMetadata,
-        referrersEnabled: settingMetadata,
-      },
-    };
+
+    apiMetadata.privacy = {};
+    for (const type of ["network", "services", "websites"]) {
+      Object.defineProperty(apiMetadata.privacy, type, {
```
Rob--W: Add `configurable: true,` to the properties of the parameter to `Object.defineProperty`, to ensure that the property can be replaced later.

See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty

src/browser-polyfill.js
```diff
@@ -491,19 +491,24 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
...
+          if (origPrivacyType) {
+            const privacyType = {};
+            for (const privacyName of Object.keys(origPrivacyType)) {
+              privacyType[privacyName] = settingMetadata;
+            }
+            apiMetadata.privacy[type] = privacyType;
+            return privacyType;
+          }
+        },
+        set: (value) => value,
```
Rob--W: This is only used internally; no setter necessary.

Manvel: When I don't use setters, [getters only error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Getter_only) is thrown when [caching the value](https://github.com/mozilla/webextension-polyfill/pull/205/files#diff-a8dc0200482fd536ca439ed8bcf763c2R505) as the `strict mode` is used.

Not sure why this is not being caught by tests.

**Steps to reproduce:**
1. `npm run build`.
2. Load `dist/browser-polyfill.js` into a chrome extension.
3. execute `browser.privacy.services.passwordSavingEnabled.get({});`

**Observer behavior:**
```
Uncaught TypeError: Cannot set property services of #<Object> which has only a getter
```

**Expected:**
No error is thrown.

src/browser-polyfill.js
Manvel: done.

src/browser-polyfill.js
Rob--W: This is because you're triggering the setter (at line 505).

There are two ways to replace a getter property:

1. `delete` the original property and assign the value.
2. Use Object.defineProperty` to replace the property descriptor with one that only has a value.

In a unit test, you should count the number of accesses to `chrome.privacy.network` when you dereference the `chrome.privacy.network` object.

After looking at the implementation, it is apparent that the metadata is dereferenced only once and cached afterwards. I guess that you don't even have to replace the getter, so the code can be simplified.

implementation where metadata is dereferenced once:

https://github.com/mozilla/webextension-polyfill/blob/5f7a8458c9b249052816f520e4f8b13bf6337000/src/browser-polyfill.js#L256-L261

Manvel: > In a unit test, you should count the number of accesses to chrome.privacy.network when you dereference the chrome.privacy.network object.

I don't understand this comment, you want me to update the tests in a way?

> After looking at the implementation, it is apparent that the metadata is dereferenced only once and cached afterwards. I guess that you don't even have to replace the getter, so the code can be simplified

Sounds good, then I guess I can just remove the line where I overwrite the getter and that's it?

Manvel: > After looking at the implementation, it is apparent that the metadata is dereferenced only once and cached afterwards. I guess that you don't even have to replace the getter, so the code can be simplified

Done, removed the cache. Hope I understood your comment correctly.

> In a unit test, you should count the number of accesses to chrome.privacy.network when you dereference the chrome.privacy.network object.

I need a hint here, not sure how I can implement that.

Manvel: @Rob--W I have added a test, looking into existing implementation you provided, but I'm bit unsure about it.

> It would probably also be good to verify that if fakeChrome.privacy.services is undefined, that window.privacy.network.services is undefined as well

Not sure if I understood this one, are you referring to something like?
```
equal(window.browser.privacy.services, undefined);
```

test/test-proxied-properties.js
```diff
@@ -178,6 +189,12 @@ describe("browser-polyfill", () => {
 
         window.chrome.runtime.onMessage.addListener(() => {});
         equal(lazyInitCount, 1, "chrome.runtime should be re-used upon accessing chrome.runtime");
+
+        window.browser.privacy.network.networkPredictionEnabled();
```
Rob--W: Don't call the function, just check that the returned object looks like a Setting object (object with set, get and clear methods).

test/test-proxied-properties.js
```diff
@@ -162,6 +163,16 @@ describe("browser-polyfill", () => {
         get tabs() {
           ok(false, "chrome.tabs should not lazily be initialized without explicit API call");
         },
+        privacy: {
+          get network() {
+            ++lazyInitPrivacyCount;
+
+            const networkPredictionEnabled = () => true;
+            const value = {networkPredictionEnabled};
+            Object.defineProperty(fakeChrome.privacy, "network", {value});
```
Rob--W: Don't replace the property. Now we're interested in how often `privacy.network` is dereferenced.

test/test-proxied-properties.js
```diff
@@ -140,6 +140,7 @@ describe("browser-polyfill", () => {
   describe("without side effects", () => {
     it("should proxy non-wrapped methods", () => {
       let lazyInitCount = 0;
+      let lazyInitPrivacyCount = 0;
```
Rob--W: Could you create a new `describe(...)` block to test the privacy API?

Manvel: Done.

test/test-proxied-properties.js
Manvel: Noted, done.

It fails, but I can confirm it's not a regression as it was calling twice for the old implementation as well.

Any idea why it's so? Might be a problem with my test?

test/test-proxied-properties.js
Manvel: Done.

test/test-proxied-properties.js
Manvel: Seems like if I replace getter in [the implementation](https://github.com/mozilla/webextension-polyfill/pull/205/files#diff-a8dc0200482fd536ca439ed8bcf763c2R505) the object will be dereferenced once.

Will do so if no objection.

Manvel: Nevermind, it doesn't help.

Manvel: > It fails, but I can confirm it's not a regression as it was calling twice for the old implementation as well.

This was wrong assumption, apparently I was running test without running `npm run build` first, Sorry for this, will have a look into it later today, the old tests were passing correctly, the problem is in the new implementation.

Manvel: After taking a closer look into implementation, now I see what is the problem.

The problem is that I can't fix it without touching `wrapAPIs()` code, unless I hardcode everything.

I'm thinking about something like:

```js
const settingMetadata = {
      clear: {minArgs: 1, maxArgs: 1},
      get: {minArgs: 1, maxArgs: 1},
      set: {minArgs: 1, maxArgs: 1},
};

apiMetadata.privacy = {"*": {"*": settingMetadata}};
```

and updating `wrapObject()`:

```
if (hasOwnProperty(metadata, "*")) {
 value = wrapObject(value, wrappers[prop], metadata["*"]);
}
```

I'm quite uncomfortable updating `wrapObject()`, but I'll give it a shot and revert it if you say so.

Manvel: @Rob--W Done.

src/browser-polyfill.js
```diff
@@ -491,19 +494,7 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
...
-      },
-      services: {
-        passwordSavingEnabled: settingMetadata,
-      },
-      websites: {
-        hyperlinkAuditingEnabled: settingMetadata,
-        referrersEnabled: settingMetadata,
-      },
-    };
+    apiMetadata.privacy = {"*": {"*": settingMetadata}};
```
Rob--W: I'm not immediately against this implementation, since it is still somewhat readable and concise.
I do wonder whether we should explicitly spell out `network`, `services`, and `websites`.

@rpl Thoughts?

Manvel: @Rob--W Any thought on this? Happy to explicitly spell out `network`, `services`, and `websites`, if that holds the progress.

CC: @rpl

rpl: :+1: from me on spelling out the first level and use the "*" to cover all the properties inside those.

test/test-proxied-properties.js
```diff
@@ -181,4 +181,35 @@ describe("browser-polyfill", () => {
...
+              get: () => {},
+              set: () => {},
+              clear: () => {},
+            };
+            return {networkPredictionEnabled};
+          },
+        },
+      };
+
+      return setupTestDOMWindow(fakeChrome).then(window => {
```
Rob--W: `equal(lazyInitCount, 0, ...);` to check that `chrome.privacy` is not accessed at first.

src/browser-polyfill.js
Manvel: Thanks @rpl and @Rob--W, done.

I was thinking about 2 different implementation:
```
const privacyType = {"*": settingMetadata};
apiMetadata.privacy = {
      network: privacyType,
      services: privacyType,
      websites: privacyType,
};
```

```
apiMetadata.privacy = {};
for (const type of ["network", "services", "websites"]) {
  apiMetadata.privacy[type] = {"*": settingMetadata};
}
```

If you have preference for one or another please let me know which to use. This seem to be most consistent comparing to `settingMetadata`.

test/test-proxied-properties.js
Manvel: Done.

src/browser-polyfill.js
Rob--W: Keep the current code. It is more readable than the proposed alternatives.

Manvel:
@Rob--W Is there anything blocking this issue to be merged into master?

Thanks in advance.

Rob--W:
Thanks for the patch @Manvel !

PR #212, opened 2019-11-16
Add missing alarms.create

evilpie:
Not Tested!

Found via https://stackoverflow.com/questions/58887376/browser-alarms-is-undefined-using-webextension-polyfill

### DIFF
--- a/api-metadata.json
+++ b/api-metadata.json
@@ -8,6 +8,10 @@
       "minArgs": 0,
       "maxArgs": 0
     },
+    "create": {
+      "minArgs": 0,
+      "maxArgs": 2
+    },
     "get": {
       "minArgs": 0,
       "maxArgs": 1


api-metadata.json
```diff
@@ -8,6 +8,10 @@
       "minArgs": 0,
       "maxArgs": 0
     },
+    "create": {
+      "minArgs": 0,
+      "maxArgs": 2
```
Rob--W: Add `"fallbackToNoCallback": true`.

Apparently `chrome.alarms.create` doesn't take a callback - https://developer.chrome.com/extensions/alarms#method-create

Rob--W:
Firefox's `browser.alarms.create` does not return a promise either.

The `alarms.create` API schedules a task, and in practice any other call in the `alarms` namespace will be ordered relative to each other. So there is no need to make the `alarms.create` API async either.

So we'll close this PR, so that both Firefox and the polyfill behave identically.

FYI, even without declaring `alarms.create` in the api-metadata.json file, the `alarms.create` method will still be available (except it doesn't return a promise when called, unless the `chrome.alarms.create` API somehow starts to return a promise).

PR #213, opened 2019-11-26
chore: Update npm dev dependencies (besides jsdom, and the ones related to babel)

rpl:
This PR covers the npm dependencies that are outdated and can be updated without any actual change, or with just a very small amount of changes (like the ones applied to update sinon to version 7.5.0).

The following outdated deps are not being updated yet in this PR:
```
Package              Current  Wanted  Latest  Location
babel-preset-babili   0.0.10  0.0.10   0.1.4  webextension-polyfill
grunt-babel            6.0.0   6.0.0   8.0.0  webextension-polyfill
jsdom                 9.12.0  9.12.0  15.2.1  webextension-polyfill
```

jsdom 15.2.1 is not yet updated in this PR because there are backward incompatible changes to the jsdom APIs we use in our test harness, I'm going to take care of that in a separate PR that also port that parts to the new jsdom API.

grunt-babel 8.0.0 isn't yet updated because depends from babel 7, and so I'm deferring it to a separate PRs (especially because it turned out that, while updating this repo to babel 7 is simple enough, there are some differences in how @babel/transform-umd-modules wraps the module which are not working as in the previous version and the browser API object isn't being exported as it should and all the tests are failing, as expected given that the UMD module isn't being generated correctly)

babel-preset-babili has not been updated because it has been more recently renamed to babel-preset-minify in more recent version, but it requires to also update the other babel dependencies and it would make sense to update to babel-preset-minify along with updating babel.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "selenium-webdriver": "^4.0.0-alpha.1",
     "serve-static": "^1.13.1",
     "shelljs": "^0.8.2",
-    "sinon": "^1.17.6",
+    "sinon": "^7.5.0",
     "tap-nirvana": "^1.0.8",
     "tape": "^4.9.1",
     "tape-async": "^2.3.0",
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "grunt": "^1.0.1",
     "grunt-babel": "^6.0.0",
     "grunt-contrib-concat": "^1.0.1",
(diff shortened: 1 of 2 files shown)


rpl:
@Rob--W the babel part is #214, currently marked as "state: not merge" and "state: in progress" because of the issue with @babel/transform-modules-umd.

Issue #202, opened 2019-09-21
Does not work when imported as a module.


MicahZoltu:
If you attempt to load the `webextension-polyfill` as a module, you get a runtime failure in both Chrome and Firefox due to the utilization of `this` as the `global` parameter.  `this` is undefined in the context of an ES module, so when the root UMD module function is called and `this` is passed to the `global` parameter, it ends up passing `undefined`.  Then when `global.browser` is referenced you get a type error.

The same problem occurs with the minified source.

Error message in Firefox:
```
TypeError: global is undefined
```

### Recommendation
Use `globalThis || window || this` instead of `this` as the first parameter.  I believe any browser that supports ES modules also supports either `globalThis` or `window`, so `this` will only be hit in the rare case.

### Repro Case

```html
<!DOCTYPE html>
<html lang='en'>
	<head>
		<meta charset='utf-8'>
	</head>
	<body>
		<script type='module' src='browser-polyfill.js'></script>
	</body>
</html>
```
```json
{
	"manifest_version": 2,
	"name": "ModuleFailureRepro",
	"version": "1.0",
	"description": "Repro case for loading a polyfill as an ES module",
	"background": {
		"page": "background.html"
	}
}
```

MicahZoltu:
A quick attempt to submit a PR shows that the problem is that the build scripts appear to be turning the script into a UMD module, and the output of that script is the problem.  Fixing this will likely require either upgrading the build script tooling, or getting the bug fixed in the build script tooling.

MicahZoltu:
It appears the issue is coming from this buildscript step:
https://github.com/mozilla/webextension-polyfill/blob/04a9e3a2234ed61b0bd77caa7f66dc50214c9192/Gruntfile.js#L75-L80

MicahZoltu:
I have [filed an issue](https://github.com/babel/babel/issues/10476) with Babel about this, TBD whether anything will come from that.

rpl:
@MicahZoltu We (@Rob--W and I) discussed about this and we are open to fix this on the polyfill side by building a separate .mjs version that is going to be a pure ES6 module (and make it part of the npm package release).

MicahZoltu:
FWIW https://github.com/babel/babel/pull/10477 has been merged.  I believe it will be available in Babel 7.7.0, whenever that release gets cut.

rpl:
@MicahZoltu that's great (also thanks for filing it upstream).

[...]

Issue #211, opened 2019-11-13
Update repo to use a more recent version of babel dependencies


rpl:
This repo is currently using a pretty old babel version, we should migrate the project to babel 7.

Migrating to Babel 7.7.0 (once released) should also fix #202.

rpl:
Closed by #214

PR #214, opened 2019-11-26
fix: Update babel deps to babel 7

rpl:
This PR updates babel to babel 7 (#211) and some other babel-related dependency (like babel-preset-babily -> babel-preset-minify, grunt-babel to version 8.0.0).

Quite unfortunately, this PR isn't ready yet because of an issue with the new version of the babel's transform-modules-umd plugin (which was one of the reasons why we wanted to update babel soon enough):

Even if the build process is completing successfully with the new version of the transform-modules-umd, our js file isn't being wrapped correctly as it was with previous versions of this babel transform plugins (and all the tests are failing, which is expected given that the wrapped module fails to export the browser API object).

In particular it seems that this babel plugin is now trying to discover the exports by parsing the module sources and it ends to don't find any and not passing any module object to the function that wraps our module source, here is an extract of the diff between the last version of the umd modules as released on npm and the one built from this branch:

```
--- ../browser-polyfill-umd-0.5.0.js    1985-10-26 09:15:00.000000000 +0100
+++ dist/browser-polyfill.js    2019-11-26 16:52:29.146299551 +0100
@@ -1,19 +1,22 @@
 (function (global, factory) {
   if (typeof define === "function" && define.amd) {
-    define("webextension-polyfill", ["module"], factory);
+    define("webextension-polyfill", [], factory);
   } else if (typeof exports !== "undefined") {
-    factory(module);
+    factory();
   } else {
     var mod = {
       exports: {}
     };
-    factory(mod);
+    factory();
     global.browser = mod.exports;
   }
-})(this, function (module) {
+})(typeof globalThis !== "undefined" ? globalThis : typeof self !== "undefined" ? self : this, function () {
```

I took a look to the [trasform module sources](https://github.com/babel/babel/blob/master/packages/babel-plugin-transform-modules-umd/src/index.js) and at a first glance it seems that it may be easier to actually replace the current strategy with a custom "less generic" transformation babel plugin (which we could write to specifically wrap our module correctly), or from a grunt task that would just wrap the module into a UMD template provided by us.


rpl:
I've looked a bit more into how the new version of the transform-umd-modules works internally and, as I assumed, explicitly including an export statement in the polyfill sources makes the transform plugin to detect the exports, e.g.:

```js
let extensionAPIs;
...
   // The build process adds a UMD wrapper around this file, which makes the
   // `module` variable available.
   extensionAPIs = wrapAPIs(chrome);
 } else {
   extensionAPIs = browser;
 }

export default {browser: extensionAPIs};
```

but the result is still not what we would like it to be, in particular because (when loaded as a regular script on Firefox) it is going to overwrite the original browser global with the default export object instead of the original browser object itself:

```js
...
    var mod = {
      exports: {}
    };
    factory(mod.exports);
    global.browser = mod.exports;
...
    extensionAPIs = wrapAPIs(chrome);
  } else {
    extensionAPIs = browser;
  }

  var _default = {
    browser: extensionAPIs
  };
  _exports.default = _default;
});
```

And so I looked into the alternative strategy mentioned in the pull request description:

a custom "less generic" transformation babel plugin (which we could write to specifically wrap our module as we need it to be)

and updated this pull request accordingly.

(and I also added a smoke test to verify as part of the automated tests that the bundle generated with this new minimal custom transformer plugin is also going to fix #202, on both Chrome and Firefox).

test/fixtures/import-as-es6-extension/background.js
```diff
@@ -0,0 +1,3 @@
+browser.runtime.onMessage.addListener(async (msg, sender) => {
```
Rob--W: The `browser` here is the global `browser` object, which is not the same as the one from the module (unless the module replaced the `browser` global).

If you want to verify that module loading works as expected, then `globalThis.browser` should be undefined (in Chrome) and defined in Firefox, and the file should start with `import {browser} from './browser-polyfill.js';`

Can you confirm that this (i.e. module script replacing the `browser` global instead of only exporting a browser object) is the intended behavior?

rpl: By looking to what the umd wrapper is doing (there are fragments of its content in the PR description), you may notice that it is passing as the `global` parameter the result of `typeof globalThis !== "undefined" ? globalThis : typeof self !== "undefined" ? self : this` and so where globalThis is defined (and the script isn't loaded as an AMD or CommonJS module) the result is that it is going to call `globalThis.browser = mod.exports`.

And so, besides exporting browser from the ES6 module, it is also exporting it as a global.
Based on what #202 describes and the fix applied upstream in babel/babel#10477, this seems to be exactly the intended behavior.

Currently the test case is mainly a smoke test for #202 expectations, but testing explicitly that `import ... from './browser-polyfill.js';` also works as expected sounds reasonable to me too, I'll look into adding an additional assertion or test case for that.

Let me know if there is anything else in this PR that you would like to discuss further, or something that may require more changes or investigation.

rpl: @Rob--W 

Actually I have to correct myself:
the generated UMD module doesn't export anything and so it can't be used as `import ... from './browser-polyfill.js'` and neither as `const ... = await import("./browser-polyfill")` because the returning module is going to be empty.

We could generate a separate "native" ES6 module which does export the API object, but I'm not sure that it is actually doable to mix the two together, because the export statement has to be on the top level scope and not inside the wrapper function scope.

I think that it may be more reasonable to defer looking more into that (generating a separate "native" ES6 module), because this is currently blocking the upgrade of this repo to a more recent babel version and I would really prefer that to be fixed as soon as we can.

The current version does pass all the existing test case, and so it should already be working as the current one, plus fixing #202 

In my opinion this is enough for now.

Rob--W: What you just explained sounds reasonable. Let's fix #202 and not worry about `browser` export for now.

PR #216, opened 2019-12-23
doc: Add README.md section related to basic setup with ES6 module loader

### LINKED ISSUE #202: Does not work when imported as a module.

rpl:
Short addition to the README file (fixes #202)

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ Table of contents
 * [Supported Browsers](#supported-browsers)
 * [Installation](#installation)
 * [Basic Setup](#basic-setup)
+  * [Basic Setup with ES6 module loader](#basic-setup-with-es6-module-loader)
   * [Basic Setup with module bundlers](#basic-setup-with-module-bundlers)
   * [Usage with webpack without bundling](#usage-with-webpack-without-bundling)
 * [Using the Promise-based APIs](#using-the-promise-based-apis)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -117,6 +118,32 @@ browser.tabs.executeScript({file: "content.js"}).then(result => {
 });
 ```
 
+### Basic Setup with ES6 module loader
+
+The polyfill can also be loaded using the native ES6 module loader available in
+the recent browsers versions.
+
+Be aware that the polyfill module does not export the `browser` API object.
```
Rob--W: Add something like this:

```
, but defines the `browser` object in the global namespace (i.e. `window`) if needed.
```

PR #220, opened 2020-02-19
chore: Remove IRC notification from Travis

Rob--W:
We moved from IRC to Matrix.

... but the project is stable with little new development so we can as well remove the notifications.


Rob--W:
I have submitted a PR to update `chromedriver` in https://github.com/mozilla/webextension-polyfill/pull/221

PR #230, opened 2020-05-17
Allowing devtools.panel.create accept up to 4 args

antfu:
In Chrome, `devtools.panel.create` accepts 4 arguments:

https://developer.chrome.com/extensions/devtools_panels#method-create

While firefox only accepts 3:

https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/devtools.panels/create

Not sure how to address this inconsistency, but IMHO the arguments shouldn't be limited to 3.

### DIFF
--- a/api-metadata.json
+++ b/api-metadata.json
@@ -214,7 +214,7 @@
     "panels": {
       "create": {
         "minArgs": 3,
-        "maxArgs": 3,
+        "maxArgs": 4,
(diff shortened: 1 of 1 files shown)


antfu:
Oh wait, I see. Callbacks are being converted to promises. Close due to my mistake 😅

PR #239, opened 2020-09-29
fix: Corrected maxArgs for topSites.get

dogagenc:
`topSites.get` method accepts an optional options argument. Although, polyfill throws an error when you invoke that method with options because maxArgs of method defined as 0 in meta-data file. This resolves problem by passing wrong validation on method argument count.

Ref: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/topSites/get

### DIFF
--- a/api-metadata.json
+++ b/api-metadata.json
@@ -606,7 +606,7 @@
   "topSites": {
     "get": {
       "minArgs": 0,
-      "maxArgs": 0
+      "maxArgs": 1
     }
   },
   "webNavigation": {


Rob--W:
Chrome does not support the options parameter, so it would throw even with `maxArgs: 1`.
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/topSites/get#Browser_compatibility

dogagenc:
@Rob--W Oh, my bad. Thank you, i'm closing this.

dogagenc:
@Rob--W To clearify it, why would it throw if it's just a validation check in polyfill? What happens if it pass validation and just invoke function with an unnecessary argument in Chrome?

Rob--W:
The polyfill wouldn't throw. It just passes on the arguments in unmodified form, upon which Chrome would throw because these parameters do not match the method's signature.

dogagenc:
@dogagenc I see, thank you. 🙏

PR #241, opened 2020-10-07
feat: Add more validation for browser apis

arthur791004:
## Goal

Fix the polyfill might not work on the website which defined browser as an object in global scope. The original behavior only test browser is an object or not. However if a website has following code in chrome, then it would return `browser` immediately which makes this polyfill not wok.

```js
window.browser = window.Object();
```

## Changes

* Check chrome apis is valid extension apis first

## Limitation

Even though checking chrome apis is valid, website could also define a dummy `chrome.runtime.id` to make this polyfill broken. IMHO, the better way might be detecting the environment (chrome, firefox or etc) directly rather than check the global variable in the future.

### DIFF
--- a/src/browser-polyfill.js
+++ b/src/browser-polyfill.js
@@ -6,7 +6,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 "use strict";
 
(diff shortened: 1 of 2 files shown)


arthur791004:
Closed because this way is not very great.

PR #247, opened 2020-10-28
feat: Added devtools.panels.elements.createSidebarPane to metadata

### LINKED ISSUE #246: browser.devtools.panels.elements.createSidebarPane('Gem') return undefined on Chrome
When I use `chrome.devtools.panels.elements.createSidebarPane` it works fine

akhilpanchal:
Fixes #246

### DIFF
--- a/api-metadata.json
+++ b/api-metadata.json
@@ -216,6 +216,12 @@
         "minArgs": 3,
         "maxArgs": 3,
         "singleCallbackArg": true
+      },
+      "elements": {
+        "createSidebarPane": {
+          "minArgs": 1,
+          "maxArgs": 2
+        }
       }
     }
   },
--- a/api-metadata.json
+++ b/api-metadata.json
@@ -220,7 +220,7 @@
       "elements": {
         "createSidebarPane": {
           "minArgs": 1,
-          "maxArgs": 2
+          "maxArgs": 1
         }
       }
     }


api-metadata.json
```diff
@@ -216,6 +216,12 @@
         "minArgs": 3,
         "maxArgs": 3,
         "singleCallbackArg": true
+      },
+      "elements": {
+        "createSidebarPane": {
+          "minArgs": 1,
+          "maxArgs": 2
```
rpl, line 223: It looks that `"maxArgs"` should be set to `1` for `devtools.elements.createSidebarPane`:
- if you look to the api docs for [`devtools.panels.create` in the chrome.devtools.panels doc page](https://developer.chrome.com/extensions/devtools_panels#method-create) you can see that it does have 3 mandatory parameters + the optional callback
- the api docs for [`devtools.panels.elements.createSidebarPane` in the same doc page](https://developer.chrome.com/extensions/devtools_panels#method-ElementsPanel-createSidebarPane) shows that it does have 1 mandatory parameter (the sidebar title) + the optional callback

and so maxArgs should be set to the maximum number of parameters that the API method does expect without counting the optional callback.

The reason for that is that the corresponding promised-based API method signature shouldn't expect a callback parameter (and the polyfill will pass a callback internally to the wrapped chrome callback-based API method), e.g.:

```
chrome.devtools.panels.elements.createSidebarPane("my sidebar", (sidebarPanel) => ...);
```
translates into:
```
const sidebarPanel = await browser.devtools.panels.elements.createSidebarPane("my sidebar");
...
// or
devtools.panels.elements.createSidebarPane("my sidebar").then(sidebarPanel => ...);
```

rpl (changes requested):

api-metadata.json
akhilpanchal: Thanks again for the explanation. That makes a lot of sense.
Could you also explain what the role of `singleCallbackArg` property is?
I looked at the following lines and I am a little unsure about when to use this property.

https://github.com/mozilla/webextension-polyfill/blob/8c7be468735c9cd275fef8705803d3a3beb792dd/src/browser-polyfill.js#L95-L100

Rob--W: For context, see https://github.com/mozilla/webextension-polyfill/issues/168#issuecomment-453623762

rpl (approved):
Looks good!
Thanks @akhilpanchal

caitmuenster:
Thanks so much, @akhilpanchal! Your contribution has been added to our [recognition wiki](https://wiki.mozilla.org/Add-ons/Contribute/Recognition). 🎉

Welcome onboard! We look forward to seeing you around.

PR #250, opened 2020-12-07
fix: wrap `onRequestFinished` to use promises

### LINKED ISSUE #249: `browser.devtools.network.onRequestFinished` does not promisify `getContent()`
On Chrome, `chrome.devtools.network.onRequestFinished.addListener` emits a `chrome.devtools.network.Request` object, which includes a `getContent()` method that invokes a callback with string content and its encoding. Correct usage looks like:

```javascript
chrome.devtools.network.onRequestFinished.addListener((req) => {
  req.getContent((content, encoding) => {
    // ...
  });
});
```

With Web Extensions, `getContent()` should return a `Promise` rather than being callback based. For example:

```javascript
browser.devtools.network.onRequestFinished.addListener(async (req) => {
  const [content, encoding] = await req.getContent();
});
```

Currently however, the Web Extensions polyfill simply passes through what it receives from Chrome, so `getContent()` is still callback based.

I wasn't able to find `browser.devtools` in [the spec](https://browserext.github.io/browserext/), however [MDN does say that this method exists and does in fact return a `Promise`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/devtools.network/onRequestFinished#Parameters:~:text=arguments%3A-,request). `firefox-webext-browser` also [says this returns `Promise<object>`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/acd551e6200f3c55dee83990e5a3419a2e0b7e56/types/firefox-webext-browser/index.d.ts#L5892). I'm not totally clear if this is an official part of the spec, or a Firefox-specific API.

My questions here are:
1. Is `browser.devtools.network.onRequestFinished` actually defined in the Web Extensions spec somewhere?
1. What exactly is the return type? `Promise<object>` is pretty vague and MDN doesn't provide specifics.
    * Chrome returns a tuple of `[string, string]`, where the former is the content as a string or a base64 encoded string (if binary), and the latter is an empty string or `'base64'` (if binary). I'm not sure what the expected behavior of Web Extensions would be for binary data?

I'm happy to write up a PR to implement this if it is appropriate to add to this polyfill. I'm not totally clear on where this API stands with respect to the spec, but if you're willing to merge it, I'm happy to write it.

dgp1130:
Fixes #249.

This updates `browser.devtools.network.onRequestFinished` to emit an
object with a promisified `getContent()` property. This brings the
polyfill implementation in line with Firefox's implementation, although
documentation is still inaccurate at the moment.

Also updates some out of date documentation with `makeCallback()`.

I tested this in my own Chrome extension and it appears to work as expected.

### DIFF
--- a/src/browser-polyfill.js
+++ b/src/browser-polyfill.js
@@ -77,13 +77,17 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
      *        promise.
      * @param {function} promise.resolve
      *        The promise's resolution function.
-     * @param {function} promise.rejection
+     * @param {function} promise.reject
      *        The promise's rejection function.
      * @param {object} metadata
      *        Metadata about the wrapped method which has created the callback.
-     * @param {integer} metadata.maxResolvedArgs
-     *        The maximum number of arguments which may be passed to the
-     *        callback created by the wrapped async function.
+     * @param {boolean} metadata.singleCallbackArg
+     *        Whether or not the promise is resolved with only the first
+     *        argument of the callback, alternatively an array of all the
+     *        callback arguments is resolved. By default, if the callback
+     *        function is invoked with only a single argument, that will be
+     *        resolved to the promise, while all arguments will be resolved as
+     *        an array if multiple are given.
      *
      * @returns {function}
      *        The generated callback function.
@@ -118,9 +122,13 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
      *        The maximum number of arguments which may be passed to the
      *        function. If called with more than this number of arguments, the
      *        wrapper will raise an exception.
-     * @param {integer} metadata.maxResolvedArgs
-     *        The maximum number of arguments which may be passed to the
-     *        callback created by the wrapped async function.
+     * @param {boolean} metadata.singleCallbackArg
+     *        Whether or not the promise is resolved with only the first
+     *        argument of the callback, alternatively an array of all the
+     *        callback arguments is resolved. By default, if the callback
+     *        function is invoked with only a single argument, that will be
+     *        resolved to the promise, while all arguments will be resolved as
+     *        an array if multiple are given.
      *
      * @returns {function(object, ...*)}
      *       The generated wrapper function.
@@ -345,6 +353,30 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
       },
     });
 
+    const onRequestFinishedWrappers = new DefaultWeakMap(listener => {
+      if (typeof listener !== "function") {
+        return listener;
+      }
+
+      /**
+       * Wraps an onRequestFinished listener function so that it will return a
+       * `getContent()` property which returns a `Promise` rather than using a
+       * callback API.
+       *
+       * @param {object} req
+       *        The HAR entry object representing the network request.
+       */
+      return function onRequestFinished(req) {
+        const wrappedReq = wrapObject(req, {} /* wrappers */, {
+          getContent: {
+            minArgs: 0,
+            maxArgs: 0,
+          },
+        });
+        listener(wrappedReq);
+      };
+    });
+
     // Keep track if the deprecation warning has been logged at least once.
     let loggedSendResponseDeprecationWarning = false;
 
@@ -480,6 +512,11 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
     };
 
     const staticWrappers = {
+      devtools: {
+        network: {
+          onRequestFinished: wrapEvent(onRequestFinishedWrappers),
+        },
+      },
       runtime: {
         onMessage: wrapEvent(onMessageWrappers),
         onMessageExternal: wrapEvent(onMessageWrappers),
--- a/test/test-onRequestFinished.js
+++ b/test/test-onRequestFinished.js
@@ -1,0 +1,119 @@
+"use strict";
+
+const {deepEqual, equal, ok} = require("chai").assert;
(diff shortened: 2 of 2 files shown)


dgp1130:
CI seems to be having a versioning error with Chrome that seems to be unrelated to this PR?

test/test-onRequestFinished.js
```diff
@@ -0,0 +1,79 @@
...
+        wrappedListener({
+          getContent(cb) {
+            cb("<html>...</html>", "text/html; charset=utf8");
+          },
+        });
+
+        ok(listener.calledOnce, "listener has not been called");
+
+        const req = listener.firstCall.args[0];
+        return req.getContent().then(([content, encodingOrMimeType]) => {
```
Rob--W, line 69: You should add a test to verify that the `getContent` method still accepts a callback.

Rob--W (changes requested):

Rob--W:
> CI seems to be having a versioning error with Chrome that seems to be unrelated to this PR?

I'll take care of that in #251

src/browser-polyfill.js
```diff
@@ -345,6 +349,32 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
...
+       * Wraps an onRequestFinished listener function so that it will return a
+       * `getContent()` property which returns a `Promise` rather than using a
+       * callback API.
+       *
+       * @param {object} req
+       *        The HAR entry object representing the network request.
+       */
+      return function onRequestFinished(req) {
+        listener({
+          ...req,
```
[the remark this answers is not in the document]
dgp1130: Good catch, updated to use `wrapObject()`, which I think is what you meant by your example. Also added a test which extends a prototype to confirm the behavior. Let me know if I'm misunderstanding how this should be implemented.

src/browser-polyfill.js
```diff
@@ -77,13 +77,17 @@ if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.
...
      *        The promise's resolution function.
-     * @param {function} promise.rejection
+     * @param {function} promise.reject
      *        The promise's rejection function.
      * @param {object} metadata
      *        Metadata about the wrapped method which has created the callback.
-     * @param {integer} metadata.maxResolvedArgs
-     *        The maximum number of arguments which may be passed to the
-     *        callback created by the wrapped async function.
+     * @param {boolean} metadata.singleCallbackArg
```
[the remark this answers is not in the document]
dgp1130: Done. `metadata.fallbackToNoCallback` and `metadata.noCallback` are still not documented for `wrapAsyncFunction()`, but I don't fully understand the subtleties of their usage, so I left them out for now.

test/test-onRequestFinished.js
dgp1130: Are you suggesting that `wrappedReq.getContent(([content, encodingOrMimeType]) => { /* ... */})` should still work? Adding a test for this fails and I believe that is the way `wrapAsyncFunction()` works so I don't think we would expect callback usage to work. I also think that isn't the way Firefox does this I would think supporting that isn't necessarily expected behavior for this polyfill.

Am I misunderstanding something or are there additional changes to be made to support a callback API on top of the `Promise` API?

Rob--W: Oops, never mind. We do intentionally not support callbacks in the API, per https://github.com/mozilla/webextension-polyfill#no-callback-supported-by-the-promise-based-apis-on-chrome

test/test-onRequestFinished.js
```diff
@@ -0,0 +1,122 @@
...
+
+        const requestProto = {
+          inheritedProp() {},
+        };
+        const request = {
+          getContent(cb) {
+            cb("<html>...</html>", "text/html; charset=utf8");
+          },
+        };
+        Object.setPrototypeOf(request, requestProto);
```
Rob--W, line 110: Simplify to
```javascript
const request = Object.create({
  inheritedProp: true,
  getContent(cb) {
    cb("", "");
  },
});
```

test/test-onRequestFinished.js
```diff
@@ -0,0 +1,122 @@
...
+           "devtools.network.onRequestFinished.addListener should have been called once");
+
+        const wrappedListener = fakeChromeOnRequestFinished.addListener.firstCall.args[0];
+        wrappedListener({
+          getContent(cb) {
+            cb("<html>...</html>", "text/html; charset=utf8");
+          },
+        });
+
+        ok(listener.calledOnce, "listener has not been called");
```
Rob--W, line 66: This and some other messages: make sure that the message describes the expected behavior.

`onRequestFinished has been called once` is a better description.

Rob--W (changes requested):

test/test-onRequestFinished.js
dgp1130: Done.

test/test-onRequestFinished.js
dgp1130: Done.

Rob--W (approved):
LGTM thanks. I'll merge it after the other PR that unbreaks the build is merged.

Rob--W:
The other PR got approved and merged. Could you rebase and update the PR so we can get a green CI build?

dgp1130:
Rebased and got a green CI, thanks for fixing that.

Rob--W:
Thanks for the report and patch!

caitmuenster:
Thanks so much for the patch, @dgp1130! Your contribution has been added to our [recognition wiki](https://wiki.mozilla.org/Add-ons/Contribute/Recognition). 

Take care!

PR #256, opened 2021-01-05
chore(ci): Migrate CI jobs to circleci

rpl:
- Migrate CI jobs from travis to circleci
- Migrate code coverage from coveralls to codecov (while migrating web-ext repo to circleci we did notice issues with using coveralls from circleci, and migrate to codecov is the solution we agreed on)

TODO:
- [x] figure out intermittent failure on integration tests (seems to triggered by a chromedriver error): https://app.circleci.com/pipelines/github/mozilla/webextension-polyfill/2/workflows/cc1d1eb8-2066-423d-bce6-e5600db690e0/jobs/2
  - this intermittent failure was due to chrome intermittently fail to connect to Xvfb, fixed by using Xephyr instead.
- [x] setup NPM_TOKEN on the circleci project
- [x] double-check project settings match the one used for other github repos (e.g. web-ext circleci settings)


rpl:
@muffinresearch would you mind to setup circleci for this repo? (and regenerate and set on the circleci side the npm token as we did for the other repos).

muffinresearch:
@rpl I don't have admin on this repo.

muffinresearch:
Access issues are resolved now and tokens are setup.

rpl:
Based on a quick investigation it is likely that the chromedriver intermittent issue (which did also log an `unknown error: DevToolsActivePort file doesn't exist` error) may be due to limitation related to `/dev/shm` while running inside VMs / docker containers.

Given that we do use chromedriver only to run some small integration tests (to make sure that the polyfill does behave as expected when execute inside a real chrome instance) we may just try to pass `--disable-dev-shm-usage` along with the other Chrome options when running inside a CI job (the Chrome instance shouldn't show any difference in behavior, the main reason for not using the flag is performance, but that shouldn't be a problem in this case given how small are the tests that we do run).

rpl:
> Based on a quick investigation it is likely that the chromedriver intermittent issue (which did also log an `unknown error: DevToolsActivePort file doesn't exist` error) may be due to limitation related to `/dev/shm` while running inside VMs / docker containers.

Scratch that... I investigated a bit more deeply the issue by connecting to the circleci job over SSH and it turned out that the chromedriver issue is actually due to chrome intermittently failing to connect to the Xvfb instance that the circleci browsers image variants does run by default. I tried to use Xephyr instead of Xvfb and the intermittent failures doesn't seem to happen anymore.

I'm going to tweak the circleci config a bit more to:
- install and run Xephyr (unfortunately it isn't installed by default in the circleci browsers image variant)
- run the integration tests on Xephyr
- collect chromedriver verbose logs and store the log file in the job artifacts (so that it may be easier for us to investigate chromedriver intermittent issues if they will happen again in the future)

rpl:
@Rob--W this PR is now ready for review (the intermittent failure was an issue with chrome failing intermittently to connect to Xvfb, the tests pass consistently using Xephyr and so I opted for that, the time to run the job is ~2min, around 20 seconds longer, and so it doesn't seem to make any big difference from that perspective neither).

.circleci/config.yml
```diff
@@ -0,0 +1,171 @@
...
+    description: run tests
+    steps:
+      - run:
+          name: run linting check and unit tests with coverage
+          command: npm run test-coverage
+      - store_artifacts:
+          path: coverage
+      - run:
+          name: publish coverage data
+          command: echo TODO npm run publish-coverage
```
Rob--W, line 99: TODO?

.circleci/config.yml
```diff
@@ -0,0 +1,171 @@
...
+      - run_test_bundlers
+      - run_functional_tests
+
+  # This is required to avoid a `EACCES` when running `npm install -g` (which is
+  # executed in the test suite).
+  configure_global_npm:
+    description: create custom directory for global npm installs
+    steps:
+      # NOTE: keep the following steps compatible with both bash and cmd.exe
+      - run: cd .. && mkdir .npm-global
```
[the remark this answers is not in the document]
rpl: sounds good, in the web-ext repo the `cd .. && mkdir .npm-global` form is needed to make the same step to run on both bash and cmd.exe (as the inline comment does mention), in this repo we don't run the job on windows CI workers and so we can make this change.

.circleci/config.yml
rpl: he, thanks for spotting it, I had to skip that step while I was trying the config and not yet changes coveralls to codecov and then I forgot of that todo :-P

.circleci/config.yml
```diff
@@ -0,0 +1,170 @@
...
+      - run:
+          command: npm run test-minified
+
+  run_test_bundlers:
+    description: rerun unit tests on webpack and browserify bundled files
+    steps:
+      - configure_global_npm
+      - run: npm install -g browserify webpack webpack-cli
+      - run: |
+          export PATH=$PATH:../.npm-global/bin
```
Rob--W, line 58: I hope that the use of `../` won't bite us at some point (e.g. by a changed `$PWD`). If you think that this is a concern too, consider using `$HOME` or `~` or any other fixed location.

Rob--W (approved):

.circleci/config.yml
rpl: I'm not really worried about it (hopefully these are not going to become the classic "last famous words" :-P), if the current working dir isn't the cloned repo as (configured in a reference and merged in all jobs as default base config) then most of the commands should be failing visibly.

Issue #257, opened 2021-01-23
Errors aren't instances of Error


fregante:
Run this in background.js:

```js
browser.permissions.request({origins:["<all_urls>"]})
	.catch(err => console.log(err instanceof Error, err))
```

Firefox:

<img width="741" alt="true" src="https://user-images.githubusercontent.com/1402241/105568764-81dc9700-5d01-11eb-9b2d-9cd19382040e.png">

Chrome:

<img width="732" alt="false" src="https://user-images.githubusercontent.com/1402241/105568759-74271180-5d01-11eb-98a6-46ee9999ddbf.png">

Source:

https://github.com/mozilla/webextension-polyfill/blob/716c90bca4c8c83aa9b9615e1ecd2b8f1e69d512/src/browser-polyfill.js#L97-L98

Suggested code:

```js
        if (extensionAPIs.runtime.lastError) {
          promise.reject(new Error(extensionAPIs.runtime.lastError.message));
```

Rob--W:
Why would you want an Error instance here? It would be rather meaningless (stack, type doesn't add value).

fregante:
Specifically i found out about this because I was using `alert(error)`, which printed a nice 'Error: blah blah' in Firefox but '[object Object]' in Chrome. Additionally I was under the impression that polyfills are meant to standardize a set of APIs, so if something works on Firefox it should work the same in Chrome too, unless impossible (like the stack part)

fregante:
I noticed an instance (heh) where a real Error is thrown https://github.com/mozilla/webextension-polyfill/blob/5f7a8458c9b249052816f520e4f8b13bf6337000/src/browser-polyfill.js#L457

PR #258, opened 2021-01-23
Reject with real Errors like Firefox does

fregante:
Fixes https://github.com/mozilla/webextension-polyfill/issues/257

### DIFF
--- a/test/fixtures/detect-existing-browser-api-object/content.js
+++ b/test/fixtures/detect-existing-browser-api-object/content.js
@@ -1,6 +1,6 @@
 /* global originalAPIObjects */
 
-test("browser api object in content script", (t) => {
+test("browser api object in content script", async (t) => {
   t.ok(browser && browser.runtime, "a global browser API object should be defined");
   t.ok(chrome && chrome.runtime, "a global chrome API object should be defined");
 
@@ -10,12 +10,27 @@ test("browser api object in content script", (t) => {
     // On Firefox, window is not the global object for content scripts, and so we expect window.browser to not
     // be defined.
     t.equal(window.browser, undefined, "window.browser is expected to be undefined on Firefox");
+
+    try {
+      await browser.storage.sync.set({a: 'a'.repeat(10000000)});
+      t.fail('It should throw when set an object over quota');
+    } catch (error) {
+      t.ok(error instanceof Error);
+    }
   } else {
     // Check that the polyfill has created a wrapped API namespace as expected.
     t.notEqual(browser.runtime, chrome.runtime, "browser.runtime and chrome.runtime should not be equal");
     // On chrome, window is the global object and so the polyfilled browser API should
     // be also equal to window.browser.
     t.equal(browser, window.browser, "browser and window.browser should be the same object");
+
+    chrome.storage.local.set({a: 'a'.repeat(10000000)}, () => {
+      t.ok(chrome.runtime.lastError, 'It should throw when set an object over quota');
+      t.equal(chrome.runtime.lastError.message, 'QUOTA_BYTES quota exceeded');
+      t.notOk(chrome.runtime.lastError instanceof Error);
+    });
+
+    t.end();
   }
 });
 
--- a/test/fixtures/detect-existing-browser-api-object/content.js
+++ b/test/fixtures/detect-existing-browser-api-object/content.js
@@ -13,7 +13,7 @@ test("browser api object in content script", async (t) => {
 
     try {
       await browser.storage.sync.set({a: 'a'.repeat(10000000)});
-      t.fail('It should throw when set an object over quota');
+      t.fail('It should throw when attempting to set an object over quota');
     } catch (error) {
       t.ok(error instanceof Error);
     }
@@ -25,7 +25,7 @@ test("browser api object in content script", async (t) => {
     t.equal(browser, window.browser, "browser and window.browser should be the same object");
 
     chrome.storage.local.set({a: 'a'.repeat(10000000)}, () => {
-      t.ok(chrome.runtime.lastError, 'It should throw when set an object over quota');
+      t.ok(chrome.runtime.lastError, 'It should throw when attempting to set an object over quota');
       t.equal(chrome.runtime.lastError.message, 'QUOTA_BYTES quota exceeded');
       t.notOk(chrome.runtime.lastError instanceof Error);
     });
--- a/test/test-async-functions.js
+++ b/test/test-async-functions.js
@@ -1,6 +1,6 @@
 "use strict";
 
-const {deepEqual, equal, fail, ok, throws} = require("chai").assert;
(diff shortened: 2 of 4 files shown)


test/test-async-functions.js
```diff
@@ -56,7 +56,7 @@ describe("browser-polyfill", () => {
     it("rejects the returned promise if chrome.runtime.lastError is not null", () => {
       const fakeChrome = {
         runtime: {
-          lastError: new Error("fake lastError"),
+          lastError: {message: "fake lastError"},
```
fregante, line 59: Other styles have the same format; Only this one had a new Error

test/test-async-functions.js
```diff
@@ -70,8 +70,11 @@ describe("browser-polyfill", () => {
 
         return window.browser.tabs.query({active: true}).then(
           () => fail("Expected a rejected promise"),
-          (err) => equal(err, fakeChrome.runtime.lastError,
-                         "Got the expected error in the rejected promise")
+          (err) => {
+            instanceOf(err, window.Error, "Expected the error to be an instance of Error");
+            equal(err.message, fakeChrome.runtime.lastError.message,
```
fregante, line 75:
```suggestion
            equal(err, new window.Error(fakeChrome.runtime.lastError.message),
```

I had also tried the above but it kept saying that `expected [Error: …] to equal [Error: …]` even if `err.constructor === window.Error`. Maybe it compared other properties as well, not just `.message`

Rob--W:
Thanks for the PR. We will discuss this during the next triage meeting on Thursday next week.

The change in the test that replaces the `Error` instance with a plain object looks good to me (as it matches Chrome's actual behavior).

I'm not sure if it is desirable to replace a plain error object with an `Error` instance, but I have no strong opinions on either direction.

rpl:
> Thanks for the PR. We will discuss this during the next triage meeting on Thursday next week.

We discussed about this today during our triage meeting and agreed that the issue this PR aims to fix is in-scope for the polyfill and we are ok on proceeding with this PR.

Rob will cover the reviews, but I do have one comment related to the test coverage:

This kind of changes should also be covered with an integration test, because that is how we are making sure that the behavior of the polyfill when running on Chrome is "as similar as possible" to the behavior the same scenario would have in Firefox (where the polyfill will be just a no-op), and to catch with a test failure when that is not the case anymore (e.g. because of changes introduced in new Firefox or Chrome versions).

The fixture extensions that are being used for the integration tests are in this directory:
- https://github.com/mozilla/webextension-polyfill/tree/master/test/fixtures
And then referenced in the list of integration test defined in this file:
- https://github.com/mozilla/webextension-polyfill/blob/master/test/integration/test-extensions-in-browser.js

Some other details (e.g. how to run the integration test locally on a single "browser type") are in this section of the CONTRIBUTING.md file:
- https://github.com/mozilla/webextension-polyfill/blob/master/CONTRIBUTING.md#integration-tests

fregante:
I added some tests and they pass in Chrome locally, but Firefox tests aren't running locally and they're not passing. Do you see anything wrong there?

Rob--W:
> I added some tests and they pass in Chrome locally, but Firefox tests aren't running locally and they're not passing. Do you see anything wrong there?

What are you running and what are you observing? Any logs?

test/fixtures/detect-existing-browser-api-object/content.js
```diff
@@ -10,12 +10,27 @@ test("browser api object in content script", (t) => {
     // On Firefox, window is not the global object for content scripts, and so we expect window.browser to not
     // be defined.
     t.equal(window.browser, undefined, "window.browser is expected to be undefined on Firefox");
+
+    try {
+      await browser.storage.local.set({a: 'a'.repeat(10000000)});
+      t.fail('It should throw when set an object over quota');
+    } catch (error) {
+      t.ok(error instanceof Error);
+    }
```
rpl, line 19: The test did actually run in the CI job and this is the failure logged in the CI job:

```
browser api object in content script (Firefox)
    ✔  a global browser API object should be defined
    ✔  a global chrome API object should be defined
    ✔  browser API object should not be changed on Firefox
    ✔  window.browser is expected to be undefined on Firefox
    ✖  It should throw when set an object over quota
    -------------------------------------------------
        operator: deepEqual
        diff: 
        source: assert@moz-extension://40d035e9-8992-4e15-a880-d7950b4594e9/tape.js:8480:54
        bound@moz-extension://40d035e9-8992-4e15-a880-d7950b4594e9/tape.js:8332:32
        [72]</</</Test.prototype.fail@moz-extension://40d035e9-8992-4e15-a880-d7950b4594e9/tape.js:8574:10
        bound@moz-extension://40d035e9-8992-4e15-a880-d7950b4594e9/tape.js:8332:32
        @moz-extension://40d035e9-8992-4e15-a880-d7950b4594e9/content.js:16:9
```

It is failing because on Firefox we don't enforce the quota for storage.local in the same way that Chrome does, but we do enforce the quota (and same limits) for the storage.sync API and so changing this to use `chrome.storage.sync.set` should make the test to pass also on Firefox.

I did run it locally as well using the command suggested in the doc page I linked before:
```
TEST_BROWSER_TYPE=firefox npm run test-integration | ./node_modules/.bin/tap-nirvana
```

isn't that working for you? any useful output to understand why?

rpl (changes requested):
As a side note, the CI job didn't run the test on chrome because it is complaining that the chromedriver version used isn't compatible with the chrome version used in the CI workers.

I see that in your branch the chromedriver version is 87, while on master is 88 (and I'm going to merge the version 89 in a minute), would you mind to rebase the PR as part of the next update on this PR so that the test will be able to run on both the browser also in the CI job?

rpl:
@fregante I'm not sure way, but the last update pushed to this PR didn't trigger the circle CI job.
It may have been a temporary issue, I don't see any errors on the circle CI side, and I tried to ask renovatebot to rebase another PR and the CI job did start as expected.

would you mind to force push again to see if that triggers the CI job as expected?

fregante:
I rebased the branch on `master`, still nothing :(

rpl:
@fregante bah, I really can't see what is going on with circle ci with your branch (nothing in it should prevent circle ci from running a job for it). 

I just tried to pull your branch locally, rebased it, squashed the fixup on the first patch and then created a PR from that branch pushed in my fork (#293)... and that (for the Murphy's law) did work just fine... sigh

Do you want to try to pull the branch from #293 and force push it in yours or recreate a new PR from it?
otherwise we may just ask @Rob--W to do a final review pass on #293, if there are no other changes needed we may just merge that one.

Let me know wdyt (and my apologies on behalf of circle ci ;-))

fregante:
Feel free to merge your PR. I tried deleting my fork and re-sending another PR, still nothing

rpl:
@fregante I'm sorry, I can't really think of any reasons why it should trigger a build (especially given that it did built it before and I have even double-checked that the third party PR are enabled on the CI side).

I'll put #293 up for review from Rob.

PR #292, opened 2021-03-10
Reject with real Errors like Firefox does

fregante:
Fixes https://github.com/mozilla/webextension-polyfill/issues/257
Replaces #258 because CI is  not working there.


fregante:
Nothing. CI hates my branches

PR #293, opened 2021-03-11
fix: Reject with real Error instances like Firefox does

### LINKED ISSUE #257: Errors aren't instances of Error

rpl:
Same as #258, re-created to double-check if circle CI would run the CI job as expected. 

Fixes #257

### DIFF
--- a/test/fixtures/detect-existing-browser-api-object/content.js
+++ b/test/fixtures/detect-existing-browser-api-object/content.js
@@ -1,6 +1,6 @@
 /* global originalAPIObjects */
 
-test("browser api object in content script", (t) => {
+test("browser api object in content script", async (t) => {
   t.ok(browser && browser.runtime, "a global browser API object should be defined");
   t.ok(chrome && chrome.runtime, "a global chrome API object should be defined");
 
@@ -10,12 +10,27 @@ test("browser api object in content script", (t) => {
     // On Firefox, window is not the global object for content scripts, and so we expect window.browser to not
     // be defined.
     t.equal(window.browser, undefined, "window.browser is expected to be undefined on Firefox");
+
+    try {
+      await browser.storage.sync.set({a: 'a'.repeat(10000000)});
+      t.fail('It should throw when attempting to set an object over quota');
+    } catch (error) {
+      t.ok(error instanceof Error);
+    }
   } else {
     // Check that the polyfill has created a wrapped API namespace as expected.
     t.notEqual(browser.runtime, chrome.runtime, "browser.runtime and chrome.runtime should not be equal");
     // On chrome, window is the global object and so the polyfilled browser API should
     // be also equal to window.browser.
     t.equal(browser, window.browser, "browser and window.browser should be the same object");
+
+    chrome.storage.local.set({a: 'a'.repeat(10000000)}, () => {
+      t.ok(chrome.runtime.lastError, 'It should throw when attempting to set an object over quota');
+      t.equal(chrome.runtime.lastError.message, 'QUOTA_BYTES quota exceeded');
+      t.notOk(chrome.runtime.lastError instanceof Error);
+    });
+
+    t.end();
   }
 });
 
--- a/test/fixtures/detect-existing-browser-api-object/content.js
+++ b/test/fixtures/detect-existing-browser-api-object/content.js
@@ -42,7 +42,7 @@ test("error types", async (t) => {
       await browser.storage.sync.set({a: 'a'.repeat(10000000)});
       t.fail('It should throw when attempting to set an object over quota');
     } catch (error) {
-      t.equal(error.message, 'QUOTA_BYTES quota exceeded');
+      t.equal(error.message, 'The storage API will not work with a temporary addon ID. Please add an explicit addon ID to your manifest. For more information see https://mzl.la/3lPk1aE.');
       t.ok(error instanceof Error);
     }
   } else {
--- a/test/fixtures/detect-existing-browser-api-object/content.js
+++ b/test/fixtures/detect-existing-browser-api-object/content.js
@@ -38,7 +38,7 @@ test("error types", async (t) => {
   if (navigator.userAgent.includes("Firefox/")) {
     try {
       await browser.storage.sync.set({a: 'a'});
-      t.fail('It should throw when attempting to set an object over quota');
+      t.fail('It should throw when attempting to call storage.sync with a temporary addon ID');
     } catch (error) {
       t.equal(error.message, 'The storage API will not work with a temporary addon ID. Please add an explicit addon ID to your manifest. For more information see https://mzl.la/3lPk1aE.');
       t.ok(error instanceof Error);
--- a/test/fixtures/detect-existing-browser-api-object/content.js
+++ b/test/fixtures/detect-existing-browser-api-object/content.js
@@ -37,7 +37,7 @@ test("browser api object in background page", async (t) => {
 test("error types", async (t) => {
   if (navigator.userAgent.includes("Firefox/")) {
     try {
-      await browser.storage.sync.set({a: 'a'.repeat(10000000)});
+      await browser.storage.sync.set({a: 'a'});
       t.fail('It should throw when attempting to set an object over quota');
     } catch (error) {
       t.equal(error.message, 'The storage API will not work with a temporary addon ID. Please add an explicit addon ID to your manifest. For more information see https://mzl.la/3lPk1aE.');
--- a/test/fixtures/detect-existing-browser-api-object/content.js
+++ b/test/fixtures/detect-existing-browser-api-object/content.js
@@ -36,7 +36,6 @@ test("browser api object in background page", async (t) => {
 
 test("error types", async (t) => {
   if (navigator.userAgent.includes("Firefox/")) {
-    t.plan(2);
     try {
       await browser.storage.sync.set({a: 'a'.repeat(10000000)});
       t.fail('It should throw when attempting to set an object over quota');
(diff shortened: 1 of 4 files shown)


rpl:
@Rob--W would you mind to continue your review in this PR? 
we couldn't figure out why CircleCIi is refusing to run the CI job on #258, and on the other PR that @fregante did try to open to workaround the issue in #258.

test/fixtures/detect-existing-browser-api-object/content.js
```diff
@@ -10,12 +10,27 @@ test("browser api object in content script", (t) => {
...
     // be also equal to window.browser.
     t.equal(browser, window.browser, "browser and window.browser should be the same object");
+
+    chrome.storage.local.set({a: 'a'.repeat(10000000)}, () => {
+      t.ok(chrome.runtime.lastError, 'It should throw when attempting to set an object over quota');
+      t.equal(chrome.runtime.lastError.message, 'QUOTA_BYTES quota exceeded');
+      t.notOk(chrome.runtime.lastError instanceof Error);
+    });
+
+    t.end();
```
Rob--W, line 33: Does this `t.end();` call cause the test runner to pause until the callback for `chrome.storage.local.set` has been called?

test/fixtures/detect-existing-browser-api-object/content.js
```diff
@@ -10,12 +10,27 @@ test("browser api object in content script", (t) => {
     // On Firefox, window is not the global object for content scripts, and so we expect window.browser to not
     // be defined.
     t.equal(window.browser, undefined, "window.browser is expected to be undefined on Firefox");
+
+    try {
```
Rob--W, line 14: These tests don't really belong to this test. Could you put them in their own location?

test/fixtures/detect-existing-browser-api-object/content.js
```diff
@@ -10,12 +10,27 @@ test("browser api object in content script", (t) => {
...
+      await browser.storage.sync.set({a: 'a'.repeat(10000000)});
+      t.fail('It should throw when attempting to set an object over quota');
+    } catch (error) {
+      t.ok(error instanceof Error);
```
Rob--W, line 18: Is it possible to check the error message? This condition is true for almost every error, including "not defined" etc.

Rob--W (changes requested):

test/fixtures/detect-existing-browser-api-object/content.js
fregante: Same file, different `test()`, right?

test/fixtures/detect-existing-browser-api-object/content.js
fregante: I used t.plan in https://github.com/mozilla/webextension-polyfill/commit/f6aece9aea78ff4d1b9a7f9d41d6f8ef54fc5ff0 to ensure the number of assertions is the same for both browsers

test/fixtures/detect-existing-browser-api-object/content.js
fregante: Done in https://github.com/mozilla/webextension-polyfill/commit/f6aece9aea78ff4d1b9a7f9d41d6f8ef54fc5ff0

fregante:
@rpl can you pick https://github.com/mozilla/webextension-polyfill/commit/f6aece9aea78ff4d1b9a7f9d41d6f8ef54fc5ff0 or give me push access to your fork?

rpl:
> @rpl can you pick [f6aece9](https://github.com/mozilla/webextension-polyfill/commit/f6aece9aea78ff4d1b9a7f9d41d6f8ef54fc5ff0) or give me push access to your fork?

just saw this and I did cherry pick f6aece9 into this branch (and I did also invite you in my fork, to make this easier if you'll need to push more changes into this branch).

rpl:
@fregante the test case is currently failing because at the moment the actual Error message in Firefox is not the same as the error in Chrome:

```
 error types (Firefox)
    ✖  should be equal
    -------------------
        operator: deepEqual
        diff: "'QUOTA_BYTES quota exceeded'" => "'The storage API will not work with a temporary addon ID. Please add an explicit addon ID to your manifest. For more information see https://mzl.la/3lPk1aE.'"
        source: assert@moz-extension://34eb5c09-f75b-46a0-84d4-49dd697a71da/tape.js:9607:54
        bound@moz-extension://34eb5c09-f75b-46a0-84d4-49dd697a71da/tape.js:9459:32
        equal@moz-extension://34eb5c09-f75b-46a0-84d4-49dd697a71da/tape.js:9768:10
        bound@moz-extension://34eb5c09-f75b-46a0-84d4-49dd697a71da/tape.js:9459:32
        @moz-extension://34eb5c09-f75b-46a0-84d4-49dd697a71da/content.js:45:9
```

We are mainly interested that the error rejected is instanceof Error that not as much as we are interested into the two error messages being exactly the same, I see that Rob asked to assert the error message in the previous review round and so I propose to just use a different expected error message for Chrome and Firefox while asserting the expected error message.

We could also add the addon id in the test fixture extension manifest (by adding the applications.gecko.id property to the manifest), but the error message will still be slightly different form the one raised on Chrome, in Firefox the error message should be this one: ["QuotaExceededError: storage.sync API call exceeded its quota limitations."](https://searchfox.org/mozilla-central/rev/aa9a7136835deb0eeba00c62bb50a4a0e2cdea2d/toolkit/components/extensions/ExtensionStorageSync.jsm)

and so I think that for the purpose of the test we could also just opt to expect the "The storage API will not work with a temporary addon ID..." error message and don't add an explicit addon id in the test extension manifest.

fregante:
I think I know what's happening with CircleCI 😂 

I registered to be able to see the results (it doesn't even let you look without it) but then I removed the CircleCI from my GitHub account.

So probably they aren't building my commits out of spite.

fregante:
> and so I think that for the purpose of the test we could also just opt to expect the "The storage API will not work with a temporary addon ID..." error message and don't add an explicit addon id in the test extension manifest.

Good idea. I'll try fixing the CI issue and push that fix.

rpl:
> I think I know what's happening with CircleCI joy
> 
> I registered to be able to see the results (it doesn't even let you look without it) but then I removed the CircleCI from my GitHub account.
> 
> So probably they aren't building my commits out of spite.

ah I see, :facepalm: .... that isn't a nice behavior from circleci, it didn't even show anything to let us understand that

rpl:
> > and so I think that for the purpose of the test we could also just opt to expect the "The storage API will not work with a temporary addon ID..." error message and don't add an explicit addon id in the test extension manifest.
> 
> Good idea. I'll try fixing the CI issue and push that fix.

thank you! ping me if CI is still refusing to run CI jobs on your commits and I'll force update this branch as a last resort to force it

test/fixtures/detect-existing-browser-api-object/content.js
```diff
@@ -33,3 +33,22 @@ test("browser api object in background page", async (t) => {
     t.ok(!reply.windowBrowserIsUnchanged, "window.browser API object should have been defined by the polyfill");
   }
 });
+
+test("error types", async (t) => {
+  if (navigator.userAgent.includes("Firefox/")) {
+    try {
+      await browser.storage.sync.set({a: 'a'.repeat(10000000)});
```
Rob--W, line 40: Based on the error message, it seems that you can remove the keys themselves. The API call is rejected without even trying to store the data.

Rob--W (approved):
r+wc

test/fixtures/detect-existing-browser-api-object/content.js
```diff
@@ -33,3 +33,22 @@ test("browser api object in background page", async (t) => {
     t.ok(!reply.windowBrowserIsUnchanged, "window.browser API object should have been defined by the polyfill");
   }
 });
+
+test("error types", async (t) => {
+  if (navigator.userAgent.includes("Firefox/")) {
+    try {
+      await browser.storage.sync.set({a: 'a'});
+      t.fail('It should throw when attempting to set an object over quota');
```
rpl, line 41: Nit, the assertion message doesn't match the actual error scenario now, we could change it to `t.fail('storage.sync should throw when called from an extension with a temporary id')` or something like that.

(@fregante I can make the change and update this branch if you don't get to it first, in the end it is just a small nit related to an assertion message and so I don't mind if you prefer that I make the change and then land this on master).

rpl: Minor tweak to one of the assertion messages applied in edaa539.

fregante:
images

philipp-tailor:
Great change! 🚀 

Are there plans to publish a new release? 

And, if not: Is `mozilla:master` stable? If so, I might change my dependency to pull from the branch instead.

rpl:
> Great change! rocket
> 
> Are there plans to publish a new release?

Yes, we should be publishing it soon, I have been busy with other projects and I didn't got to come back to this (I'll mainly need to review the changes that we will release, generate a changelog and agree with Rob if we should release it as a new major version, which I think it may be reasonable based on some of the changes that I recall)  
 
> And, if not: Is `mozilla:master` stable? If so, I might change my dependency to pull from the branch instead.

We run integration tests on each change and so it should be fairly stable, but there may be some changes in behavior that may be worth to keep into account (well, one is this one, which you are already aware of).

popeda97 (commented):
Change

PR #345, opened 2021-10-28
chore: Fix missing codecov reports

rpl:
This PR includes an alternative approach to #344 

The resulting code coverage mapping is not yet right (in both this PR and #344), which is due to the fact that the sourcemap is generated by babel while converting the script into an UMD module, but after grunt-replace is used to interpolated the api-metadata.json data into the script.

I've verified locally that by manually replacing the content of api-metadata.json into the source file, than this PR does generate the expected mapping between the code coverage data collected and the original source file.

This PR is meant to be enough to ensure that codecov will detect and visibly report to us with a failure when a PR is decreasing the code coverage stats.

As a followup we could look into using babel to interpolate the api-metadata.json into the source file at an AST level using a custom babel plugin (specific to this project to keep it simple).

TODO:
- [x] verify codecov does add a code coverage comment as expected
- [x] verify that adding a fake commit that decrease the code coverage triggers a failure
- [x] file a follow up issue (#348) and link it into the inline comment added to test/setup.js

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -95,6 +95,8 @@ commands:
       - run:
           name: run linting check and unit tests with coverage
           command: npm run test-coverage
+          environment:
+            COVERAGE_WITH_SOURCEMAP: "1"
       - store_artifacts:
           path: coverage
       - codecov/upload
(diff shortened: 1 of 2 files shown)


willdurand:
> I've verified locally that by manually replacing the content of api-metadata.json into the source file, than this PR does generate the expected mapping between the code coverage data collected and the original source file.

Excellent!

willdurand (approved):
r+wc, thanks

Issue #350, opened 2021-11-09
Polyfills polyfill globals


fregante:
This polyfill is not acting as a polyfill, but rather as a regular module when used via bundlers. 

```js
import 'webextension-polyfill';
```

Polyfills are supposed to enable the API they're polyfilling; In this case, `globalThis.browser` is not available after bundling the above piece of code. That's unexpected for a polyfill. You would never `import {trimLeft} from 'string.prototype.trim-polyfill` (outside of testing environments), but rather `import 'string.prototype.trim-polyfill'`

## Possible solutions

- breaking: set `globalThis.browser` if it doesn't already exist, no UMD all 
- kinda-breaking: set `globalThis.browser` if it doesn't already exist AND set `module.exports`
- non-breaking: include 2 files:
	- the current UMD bundle (`import browser from 'webextension-polyfill'`)
	- a global-only bundle (`import 'webextension-polyfill/register'`)

or:

- meet me half way and into the future:
	- global-only bundle, for everyone even outside npm (`import 'webextension-polyfill'`)
	- ES module with `import browser from 'webextension-polyfill/module'`

## Side notes about UMD

- AMD is dead
- CommonJS is the de-facto standard, but if you really wanted to you could start distributing proper a ES module its place (on its own this would not resolve this issue though)

## Context

- https://github.com/mozilla/webextension-polyfill/issues/231
- https://github.com/mozilla/webextension-polyfill/issues/169
- https://github.com/mozilla/webextension-polyfill/issues/100
- https://github.com/mozilla/webextension-polyfill/issues/73

Rob--W:
> set `globalThis.browser` if it doesn't already exist, no UMD all

Note that `globalThis.browser` could also be set if the page has an element with `id="browser"`. That's why the current check in the polyfill looks so complicated. "if it doesn't already exist" should not be a falsey check, but at least match the condition at https://github.com/mozilla/webextension-polyfill/blob/5dbb8d94b562cc05b1c87983ba066f37c829cacc/src/browser-polyfill.js#L9. For context see https://github.com/mozilla/webextension-polyfill/pull/153

fregante:
> Note that `globalThis.browser` could also be set if the page has an element with `id="browser"`. That's why the current check in the polyfill looks so complicated. "if it doesn't already exist" should not be a falsey check, but at least match the condition at

That was a simplification, I'm just talking about dropping UMD, not the polyfill detection part.

PR #351, opened 2021-11-09
Enable usage via `ProvidePlugin`

fregante:
- Actually fixes https://github.com/mozilla/webextension-polyfill/issues/156

Rereading [my last attempt](https://github.com/webpack/webpack/issues/6226#issuecomment-720622514) at dissuading webpack developers to fix ProvidePlugin’s awkward self-providing behavior, I realized that there must be a way to fool webpack into ignoring this replacement.

```js
plugins: [
  new ProvidePlugin({
    browser: "webextension-polyfill"
  })
]
```

ProvidePlugin looks for the raw `browser` variable and replaces it; Turning it into a `window` property access should break it — if not, `typeof window["brow" + "ser"]` certainly will.

This change should be non-breaking, as an alternative solution to https://github.com/mozilla/webextension-polyfill/issues/350 — rather, as a stopgap solution to it.

### DIFF
--- a/src/browser-polyfill.js
+++ b/src/browser-polyfill.js
@@ -6,11 +6,11 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 "use strict";
 
-if (typeof window != "object" || typeof chrome != "object" || !chrome || !chrome.runtime || !chrome.runtime.id) {
+if (typeof globalThis != "object" || typeof chrome != "object" || !chrome || !chrome.runtime || !chrome.runtime.id) {
   throw new Error("This script should only be loaded in a browser extension.");
 }
 
-if (typeof window.browser === "undefined" || Object.getPrototypeOf(window.browser) !== Object.prototype) {
+if (typeof globalThis.browser === "undefined" || Object.getPrototypeOf(globalThis.browser) !== Object.prototype) {
   const CHROME_SEND_MESSAGE_CALLBACK_NO_RESPONSE_MESSAGE = "The message port closed before a response was received.";
   const SEND_RESPONSE_DEPRECATION_WARNING = "Returning a Promise is the preferred way to send a reply from an onMessage/onMessageExternal listener, as the sendResponse will be removed from the specs (See https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/API/runtime/onMessage)";
 
--- a/src/browser-polyfill.js
+++ b/src/browser-polyfill.js
@@ -6,7 +6,11 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 "use strict";
 
-if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.prototype) {
+if (typeof window != "object" || typeof chrome != "object" || !chrome || !chrome.runtime || !chrome.runtime.id) {
+  throw new Error("This script should only be loaded in a browser extension.");
+}
+
+if (typeof window.browser === "undefined" || Object.getPrototypeOf(window.browser) !== Object.prototype) {
   const CHROME_SEND_MESSAGE_CALLBACK_NO_RESPONSE_MESSAGE = "The message port closed before a response was received.";
   const SEND_RESPONSE_DEPRECATION_WARNING = "Returning a Promise is the preferred way to send a reply from an onMessage/onMessageExternal listener, as the sendResponse will be removed from the specs (See https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/API/runtime/onMessage)";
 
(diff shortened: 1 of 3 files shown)


rpl:
@fregante this change breaks the expected behavior when running on Firefox (the CI job failure is triggered by the "the browser API object should not be changed on Firefox" assertion, which was added on purpose to prevent to missing that kind of issue).

We discussed about #350 during our last triage and we agreed that building an additional version of the polyfill as a pure ESM version (in addition to the UMD one, we can evaluate to remove support for the older kinds of module loading in future version, separately from better supporting ESM module style) would be reasonable from our perspective.

would that be a reasonable solution to #350 also from your perspective?

fregante:
> this change breaks the expected behavior when running on Firefox

I don't think it does, it's more likely that the tests are broken. The conditions are unchanged except for an additional `window.browser` check, which should not affect this. 

Also the tests pass locally so I'm not sure why it's failing on CI. Could you take a look at it again? You can see that it's working correctly because the same tests passes in the background page, so the logic should be sound.

I'm not sure about what it takes to fix the tests though or why the content script tests are failing specifically.

> we agreed that building an additional version of the polyfill as a pure ESM version

🎊 



> would that be a reasonable solution to #350 also from your perspective?

#350 yes, but this PR is for #156. It only affects #350 insofar as allowing the usage via webpack as a regular polyfill.

src/browser-polyfill.js
```diff
@@ -10,7 +10,7 @@ if (typeof window != "object" || typeof chrome != "object" || !chrome || !chrome
   throw new Error("This script should only be loaded in a browser extension.");
 }
 
-if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.prototype) {
+if (typeof window.browser === "undefined" || Object.getPrototypeOf(window.browser) !== Object.prototype) {
```
fregante, line 13: This part seems to be failing the test and I'm not sure why that is. How can it pass with `browser` but not `window.browser`?

Rob--W: In Firefox's content scripts, the global is not the same as window. Conceptually you can view the global as an object (global) with its prototype set to `window`.

fregante: Interesting. What could be a solution here? `globalThis.browser` or `this.browser`?

src/browser-polyfill.js
```diff
@@ -6,7 +6,11 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 "use strict";
 
-if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.prototype) {
+if (typeof globalThis != "object" || typeof chrome != "object" || !chrome || !chrome.runtime || !chrome.runtime.id) {
```
fregante, line 9: Why is this failing now? The tests seem to be running on Chrome 90. `globalThis` was added 3 years ago in v71

rpl: @fregante I just looked to the failures in the last CI job and in the last one failed on the `npm install` step because most of the unit tests did fail.

I suspected that it was because of the nodejs version that is currently being used in the CI jobs, and that seems confirmed by the fact that `globalThis` has been introduced in nodejs 12 ([compatibility table on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis#browser_compatibility)) and the fact that in the current CI config the nodejs version is definitely set to v10.

I just opened #353 to bump the nodejs version to v12 in the CI config, as soon as I have done fighting with the chromedriver related failure I'm going to merge it. 

After that, you may rebase this PR to trigger the CI job on nodejs v12 and confirm if that was the case.

rpl: @fregante I merged #354, and so now the CI jobs from any branch based on that CI config should be using nodejs v12, and so you can give it a try.

fregante: That worked!

fregante:
I tested this [on Refined GitHub](https://github.com/refined-github/refined-github/pull/5111) and it worked in both Firefox and Chrome. 

The last commit was required to make it work in Firefox because the _free-range_ `browser` variable at the end of the file was still being replaced by ProvidePlugin.

fregante:
Should this be merged? It's been open for a while already

rpl:
> Should this be merged? It's been open for a while

yeah, sorry about that, I have not been able to get back to this repo during December, I've been trying to schedule some time to spend for some maintenance work over the github repos once I got back but didn't got to it yet.

Anyway, the change in the final version of this patch is still minimal and the existing tests are now all passing, this should be ok to merge and then release in a new webextension-polyfill version, in the meantime I'm going to approve and merge this change in a minute.

I should be able to spend some more time and cut a new release in the next few days (hopefully by the end of the week).

rpl (approved):

Issue #382, opened 2022-06-04
MV3 scripting API doesn't seem to be available


Manvel:
It seems like [MV3 scripting API](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/scripting) is not available yet. 

I am not sure what is the plan for `webextension-polyfill` to support MV3, but I am happy to try helping adding support for `scripting` API, if you say that you plan supporting MV3.

P.S. Also thanks to Mozilla team for helping devs to bring Web Extensions to more platforms with this amazing tool ❤️

Rob--W:
The behavior of the polyfill with MV3 should be documented somewhere, but for now the answer can be found in my earlier comment at #329. The behavior of the polyfill is explained at https://github.com/mozilla/webextension-polyfill/issues/329#issuecomment-884938069.

The `scripting` API supports promises already, so the `scripting` API should be usable even without changes to the polyfill.

Manvel:
Thanks for the reply @Rob--W 

> The scripting API supports promises already, so the scripting API should be usable even without changes to the polyfill.

Interesting, the last time I checked it wasn't available through `browser` namespace, but [only after scripting were added to api-metadata.json ](https://github.com/mozilla/webextension-polyfill/pull/383/files).

I will do tests again and provide minimized test case over the weekend. Probably I am missing smth. 🤔

Manvel:
Yeah, I can confirm that `scripting` still works, not sure why previously I couldn't make it work, might be missing permission. Sorry for the noise, should have tested more attentively.

PR #383, opened 2022-06-04
feat: support mv3 scripting api

Manvel:
As mentioned in https://github.com/mozilla/webextension-polyfill/issues/382  It seems like [MV3 scripting API](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/scripting) is not available yet. 

## What is included
- Added `scripting.executeScript`, `scripting.getRegisteredContentScripts`, `scripting.insertCSS`, `scripting.registerContentScripts`, `scripting.removeCSS`, `scripting.updateContentScripts` and `scripting.unregisterContentScripts` APIs to `api-metadata.json`.

## What is missing
- Tests? Not yet sure which tests would make sense here (Any pointers?)

### DIFF
--- a/api-metadata.json
+++ b/api-metadata.json
@@ -453,6 +453,36 @@
       "maxArgs": 1
     }
   },
+  "scripting": {
+    "executeScript": {
+      "minArgs": 1,
+      "maxArgs": 1
+    },
+    "getRegisteredContentScripts": {
+      "minArgs": 0,
+      "maxArgs": 1
+    },
+    "insertCSS": {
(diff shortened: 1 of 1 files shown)


Manvel:
@rpl, @Rob--W can you please have a look in case you might have a chance?

Manvel:
Closing this as of [current comment](https://github.com/mozilla/webextension-polyfill/issues/382#issuecomment-1151145505):

> The scripting API supports promises already, so the scripting API should be usable even without changes to the polyfill.

Issue #384, opened 2022-06-05
A listener indicated an asynchronous response by returning true, but the message channel closed before a response was received


fregante:
I'm seeing this message in Chrome 102 and the 5-result search page suggests that this is a new message: https://www.google.com/search?q=%22A+listener+indicated+an+asynchronous+response+by+returning+true%22

<img width="377" alt="Screen Shot 2" src="https://user-images.githubusercontent.com/1402241/172036569-7ce38b28-0de1-41cf-bd48-fc0455b36877.png">

## No-polyfill repro

```js
// in background
chrome.tabs.sendMessage(65, 'good soup', () => console.error(chrome.runtime.lastError))

// in tab
chrome.runtime.onMessage.addListener(() => true)
```

Then close the tab

## Polyfill repro

```js
// in background
await browser.tabs.sendMessage(65, 'good soup')

// in tab, never-resolving promise
browser.runtime.onMessage.addListener(() => new Promise(console.info))
```

Then close the tab

## Related

- https://github.com/mozilla/webextension-polyfill/issues/365
- https://github.com/pixiebrix/webext-messenger/issues/66

fregante:
This might actually be an opportunity to fix this very related issue:

- https://github.com/mozilla/webextension-polyfill/issues/365

But I'd still rather see a quick fix merged as soon as possible.

Rob--W:
This changed in https://bugs.chromium.org/p/chromium/issues/detail?id=1304272.
The original message was split in two different cases in https://chromium.googlesource.com/chromium/src/+/9256bc8b668f61e755ab587f623e7da94d8a6f47%5E%21/#F4, by emitting "A listener indicated an asynchronous response by returning true, but the message channel closed before a response was received" instead of "Could not establish connection. Receiving end does not exist." in some cases.

The fix in #385 looks good.

digitalmukati:
Working For Me -- 

The background script (service worker in MV3) could be going to inactive state without sending a response back to a message it received from a content script.

Example:

Background script:

chrome.runtime.onMessage.addListener((request, sender, sendResponse) => {
  // ... handle message
  return true // Error message says you already return true
})
Most MV3 APIs are asynchronous and can return promises when it makes sense to do so. (In some cases, like event listeners (e.g.: chrome.tabs.onRemoved), returning a promise wouldn't make sense). Reading a response back however can be done using callbacks or promise-style.

Content script: method 1 to read response:

[...]

PR #385, opened 2022-06-05
Handle new "tab closed" response from Chrome

fregante:
- Presumably fixes https://github.com/mozilla/webextension-polyfill/issues/384


Rob--W (changes requested):
Could you add a unit test? https://github.com/mozilla/webextension-polyfill/blob/2ece178ca125a0808faa44f2337c54588a95d20a/test/test-runtime-onMessage.js#L242-L265

If you'd like, it would also be nice to add an integration test that triggers the error in a real Chrome browser (in case the error message changes), at https://github.com/mozilla/webextension-polyfill/tree/master/test/fixtures

fregante:
Sorry I haven't had time to finish this.

PR #406, opened 2022-08-09
fix: Remove obsolete deprecation warning logged on sendResponse callback calls

### LINKED ISSUE #386: Remove the warning logged when sendResponse callback is being called
The warning ([`SEND_RESPONSE_DEPRECATION_WARNING`](https://github.com/mozilla/webextension-polyfill/blob/2ece178ca125a0808faa44f2337c54588a95d20a/src/browser-polyfill.js#L15)) has been introduced as part of #97, along with introducing support for the `sendResponse` callback (which wasn't previously supported by the polyfill at all).

The warning refers to a W3C spec drafts that was created by a different W3C community group from the one being active right now, and those W3C spec drafts are now considered obsolete.

This issue is tracking the removal of the warning from the webextension-polyfill. 
 
Some more historical context can be found in https://github.com/mozilla/webextension-polyfill/pull/97#issue-306485710

rpl:
fixes #386

### DIFF
--- a/src/browser-polyfill.js
+++ b/src/browser-polyfill.js
@@ -12,7 +12,6 @@ if (!globalThis.chrome?.runtime?.id) {
 
 if (typeof globalThis.browser === "undefined" || Object.getPrototypeOf(globalThis.browser) !== Object.prototype) {
   const CHROME_SEND_MESSAGE_CALLBACK_NO_RESPONSE_MESSAGE = "The message port closed before a response was received.";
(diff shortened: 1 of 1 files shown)


rpl:
@Rob--W I wouldn't worry about covecov complaining about the reduced coverage, that warning wasn't explicitly covered but I guess that from codecov prespective there are just less lines of code and the same number of misses and that ends up triggering its codecov/project check failure.

Rob--W (approved):

Marvintm1977au (commented):
src/browser-polyfill.js[](url)

PR #423, opened 2023-01-06
Use Dependabot instead of Renovate

### LINKED ISSUE #422: Please use Dependabot instead of Renovate
We want to use Dependabot consistently across all our (add-ons) repos. Please update the config to use Dependabot instead of Renovate. You may also want to remove the Renovate config and its GitHub integration. Thanks!

willdurand:
Fixes #422

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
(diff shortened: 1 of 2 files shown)


willdurand:
So this is failing but it is unrelated to this change.

Rob--W (approved):

PR #437, opened 2023-02-27
Update browser-tools ORB

willdurand:
This fixes an issue described in https://github.com/CircleCI-Public/browser-tools-orb/issues/62 and found by @rpl the other day.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,7 +16,7 @@ version: 2.1
 
 orbs:
   codecov: codecov/codecov@3.2.3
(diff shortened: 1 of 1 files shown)


willdurand:
I think we need https://github.com/mozilla/webextension-polyfill/pull/435 too

rpl:
> I think we need #435 too

yep, ty for tracking this one down :heart: 

I opted for cherrypicking the commit from this PR into #435, which also confirmed that with both the CI job passes as expected. Closing this one as superseded by the updated version of #435.

PR #581, opened 2023-12-26
Fix support of old browsers

maximtop:
Optional chaining operators are supported starting from the 80 version
https://caniuse.com/mdn-javascript_operators_optional_chaining
In readme is stated support for 79 version. Lets remove optional chaining operator.


Rob--W:
Fixed by #582.

PR #582, opened 2024-03-07
Fix no-op check to work with Safari and support of old browsers

xeenon:
The `browser` object is not a direct descendant of `Object.prototype` in Safari. This check should use `instanceof` for a more compatible check.

### DIFF
--- a/src/browser-polyfill.js
+++ b/src/browser-polyfill.js
@@ -6,11 +6,11 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 "use strict";
 
-if (!globalThis.chrome?.runtime?.id) {
+if (!(globalThis.chrome && globalThis.chrome.runtime && globalThis.chrome.runtime.id)) {
   throw new Error("This script should only be loaded in a browser extension.");
 }
 
-if (typeof globalThis.browser === "undefined" || Object.getPrototypeOf(globalThis.browser) !== Object.prototype) {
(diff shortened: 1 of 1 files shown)


xeenon:
@diox @rpl @willdurand @Rob--W

Rob--W:
The reason for looking up the immediate prototype is to avoid activating the polyfill when there is an element on the page that has an ID. Here is the PR that added the logic for more context: https://github.com/mozilla/webextension-polyfill/pull/153 (originally checking `instanceof Node`, but changed in response to https://github.com/mozilla/webextension-polyfill/pull/153#issuecomment-409923064 ).

What is the actual prototype chain in Safari? Is it expected to be somewhat stable?

xeenon:
> What is the actual prototype chain in Safari? Is it expected to be somewhat stable?

The problem is the prototype is not reachable, so it can't be compared with the result of `Object.getPrototypeOf()`.

I think a better approach would be to do a similar optional chain check looking for `!globalThis.browser?.runtime?.id`.

Rob--W (approved):

Rob--W:
Thanks for the patch @xeenon! We'll fix our CI and then publish a new version.

Rob--W:
We still need to publish an update FYI.

Rob--W:
FYI just released 0.11.0 that includes this patch: https://github.com/mozilla/webextension-polyfill/releases/tag/0.11.0

PR #614, opened 2024-05-14
fix: Remove webpack and webpack-cli dependencies

### LINKED ISSUE #604: Avoiding relying on WebPack in production
I'm not sure how it happened exactly, but this module seems to depend on webpack and webpack-cli, which are listed as production dependencies in the package release to NPM, even though the [`package.json`](https://github.com/mozilla/webextension-polyfill/blob/master/package.json) in the repo doesn't list any production dependencies, and depending on webpack and webpack-cli in production doesn't really make sense for almost any package.

Can this be fixed?

Gudahtt:
Changes made in #597 accidentally resulted in `webpack` and `webpack-cli` being added as dependencies in the manifest of the published package. This problem can be seen in v0.11.0. For more information, see #604.

`webpack` and `webpack-cli` are used for bundler smoke tests that are run in CI. Rather than being installed during CI, these dependencies are now included in `devDependencies`. A new npm script has been added for running this script as well, to ensure that these dependencies are on the PATH when this script is run.

The CircleCI config and the contributor docs have been updated to use the new npm script. You can test it by running
`npm run test-module-bundlers-smoketest`.

Fixes #604

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,10 +63,7 @@ commands:
   run_test_bundlers:
     description: rerun unit tests on webpack and browserify bundled files
     steps:
-      - run: npm install browserify webpack webpack-cli
-      - run: |
-          export PATH=$PATH:$PWD/node_modules/.bin
-          node ./scripts/run-module-bundlers-smoketests.js
+      - run: npm run test-module-bundlers-smoketests
 
   run_xephyr:
     description: run Xephyr on DISPLAY=:10
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,7 +33,7 @@ respectively, and finally execute the unit tests on the generated dist files.
 This project provides two test suites:
 
 - unit tests (which only require Node.js to run)
-- module bundlers smoke tests (which requires also browserify and webpack to be installed globally)
+- module bundlers smoke tests
 - integration tests (which requires also a stable version of Chrome and Firefox)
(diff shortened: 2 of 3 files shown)


.circleci/config.yml
```diff
@@ -63,10 +63,7 @@ commands:
   run_test_bundlers:
     description: rerun unit tests on webpack and browserify bundled files
     steps:
-      - run: npm install browserify webpack webpack-cli
```
Rob--W, line 66: The fix for the bug is to add `--no-save` here, as mentioned at https://github.com/mozilla/webextension-polyfill/issues/604#issuecomment-2110749236

Rob--W (approved):

Rob--W:
Thanks for the analysis and fix. Although `--no-save` could have fixed the issue, adding it to devDependencies signals more clearly that it is a potential dependency in any test. It would also make it easier to identify external regressions in the smoke tests, hence I merged this PR instead of adding a `--no-save` fix.

Gudahtt:
Thanks! Good point that `--no-save` would have been simpler, I forgot that flag existed. I prefer this fix for the same reasons you mentioned though.

.circleci/config.yml
yndaii, line 66: e2754582c72f99742aea07770b51c24ca3349f61

yndaii (changes requested):

PR #934, opened 2026-07-24
ci: bump chromedriver to 150.0.4

willdurand:


willdurand:
it makes no sense to upgrade in fact, since MV2 support has been dropped in recent chromium versions.

PR #936, opened 2026-07-24
ci: install chromium with pinned version

willdurand:


willdurand:
ok, I give up now. I don't think we can fix the chromium smoke test.

PR #938, opened 2026-07-28
ci: update browser versions

willdurand:


willdurand:
oh, ok so this in fact caused by https://developer.chrome.com/blog/extension-news-june-2025#removing_the_--load-extension_flag. Sweet.

willdurand:
OK, most tests are passing, which is an improvement over... before. r=me

